PR #3, opened 2015-08-07
Add pop() operation to Set API, to pop elements from sets easily (UNTESTED).

kno10:
Add a pop() function to the Set API.
Similar to a Stack, but without a predefined order.
The HashSets will usually pop the last occupied bucket, the tree sets return the last element.

Important: I have not yet tested all this, only done the gruntwork.
I'm not sure if I always chose the most effective solution, or if I got all the macros right.
So far, this has largely been a trial and error hack. Consider it a feature request for a "pop()" method, not a finished contribution.

My primary use case is to be able to pop integers from a hash set of integers, where I do not care about the order.

### DIFF
--- a/drv/AVLTreeMap.drv
+++ b/drv/AVLTreeMap.drv
@@ -1319,6 +1319,14 @@ public class AVL_TREE_MAP KEY_VALUE_GENERIC extends ABSTRACT_SORTED_MAP  KEY_VAL
 					return f != null;
 				}
 
+				@SuppressWarnings("unchecked")
+				public MAP.Entry KEY_VALUE_GENERIC pop() {
+					if (lastEntry == null) { throw new NoSuchElementException(); }
+					BasicEntry KEY_VALUE_GENERIC e = new BasicEntry KEY_VALUE_GENERIC ( lastEntry.key, lastEntry.value );
+					AVL_TREE_MAP.this.REMOVE_VALUE( lastEntry.key );
+					return e;
+				}
+
 				public int size() { return count; }
 				public void clear() { AVL_TREE_MAP.this.clear(); }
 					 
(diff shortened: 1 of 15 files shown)


kno10:
Did some small fixes; some iterations had the wrong starting point.

I was wondering whether it would make sense to cache the iteration point for pop() on hash sets. Because when emptying the hash set from the end, the same elements will be scanned repeatedly, so pop() would on average take n/2 steps to find the next element to return.
I'm experimenting with that right now.

vigna:
Sorry, but seems to be a bit overkill. If you do that frequently, you can use an iterator. If you do it very rarely, you can create a one-off iterator. Do you have other use cases?

kno10:
I'm needing this frequently (so I'd like to avoid creating new iterators), with insertions inbetween (so reusing the iterator will fail because of concurrent modifications).

Keeping a list of candidates (internally "int", but not exposed), polling an unspecific candidate from the set, and then maybe adding new candidates instead.

vigna:
Note that if you really need a random element, I'd do it randomly. Like, start at a random point in the backing array and return the first element.

kno10:
It does explicitly not need to be random, but "unspecific" is okay. Any element is fine, in-order is fine, too - all candidates need to be processed in arbitrary order. But because candidates can be nominated multiple times, I need to manage open and closed sets that are frequently modified.

vigna:
Well, I'd subclass in your code. Sorry, but it seems to be a bit too specific for code to end up in the distribution.

kno10:
The problem with subclassing it in my code (which I am doing in an experimental branch already) is that it is likely to break with later versions. It depends on internals of fastutil classes, so I'd prefer it to live in fastutil code.
E.g. [DBSCAN](https://en.wikipedia.org/wiki/DBSCAN) benefits from this kind of operation: iterating over a set of points to expand, and every expansion may add new candidates. A set structure is great to avoid duplicate candidates (assuming that the canidates will be only a small fragment of the database), but it needs an efficient "pop" operation.

vigna:
I understand: but for the same reason I don't want to add methods unless there are very good reasons—they must be maintained in the face of internal changes.

Really using IntOpenHashSet.iterator().next() slows down your application significantly? Do you have benchmarks?

kno10:
It is not solved in above patch yet (I just realized), but there is a subtle performance problem with repeated use of new iterators with `next()` and `remove()`.
Every call will start scanning the hash table from the end. But restarting at the end every time means that emptying this set requires n scans over on average n/2 times the hash table and thus is in O(n^2).

The code I'm using in my fastutil test branch is this:
```
    /**
     * Scan position for pop().
     */
    private transient int c = -1;

    /**
     * Pop a single value from the hash table.
     *
     * @return Value removed.
     */
    public int popInt() {
      if(size == 0) {
        throw new NoSuchElementException();
      }
      if(containsNull) {
        containsNull = false;
        --size;
        return 0;
      }
      final int key[] = this.key;
      int k, pos = c < key.length ? c : key.length;
      for(;;) {
        if (pos <= 0) pos = key.length;
        k = key[--pos];
        if(k != 0) {
          size--;
          shiftKeys(pos);
          if(size < maxFill >> 2 && n > DEFAULT_INITIAL_SIZE) {
            rehash(n >> 1);
          }
          c = pos;
          return k;
        }
      }
    }
```
which will simply continue scanning where it left off, and restart from the end until the hash table is empty. So it should reduce the run time of emptying a set via `pop()` from O(n^2) to O(n). This approach requires an additional (transient) integer for the position, and could (should?) live in a special kind of non-fail-iterator instead, I guess.

Closing the pull request, because this version is not ready for merging.

PR #4, opened 2015-08-07
Add addOrGet(k) to ObjectHashSets.

kno10:
Works like add(k) followed by get(k), but is cheaper.

This is useful for managing unique objects, e.g.

ObjectHashSet<String> unique;
key = unique.addOrGet(key);

I have wrapped this code in "#if #keyclass(Object)", because for primitives it is equivalent to "add(k); return k;".

For hash sets, the code duplication isn't too bad. But for the tree-based sets it is massive, so I didn't do these.

Again, untested code, sorry. Consider this to be a feature request.

### DIFF
--- a/drv/OpenHashBigSet.drv
+++ b/drv/OpenHashBigSet.drv
@@ -319,6 +319,46 @@ public class OPEN_HASH_BIG_SET KEY_GENERIC extends ABSTRACT_SET KEY_GENERIC impl
 		return true;
 	}
 
+#if #keyclass(Object)
+	/** Add a random element if not present, get the existing value if already present.
+	 *
+	 * This is equivalent to (but faster than) doing a:
+	 * <pre>
+	 * K exist = set.get(k);
+	 * if (exist == null) {
+	 *   set.add(k);
+	 *   exist = k;
+	 * }
+	 * </pre>
+	 */
+	public KEY_GENERIC_TYPE addOrGet( final KEY_GENERIC_TYPE k ) {
+		int displ, base;
+
+		if ( KEY_IS_NULL( k ) ) {
+			if ( containsNull ) return null;
+			containsNull = true;
+		}
+		else {
+			KEY_GENERIC_TYPE curr;
+			final KEY_GENERIC_TYPE[][] key = this.key;
+			final long h = KEY2LONGHASH( k );
+
+			// The starting point.
+			if ( ! KEY_IS_NULL( curr = key[ base = (int)( ( h & mask ) >>> BigArrays.SEGMENT_SHIFT ) ][ displ = (int)( h & segmentMask ) ] ) ) {
+				if ( KEY_EQUALS_NOT_NULL( curr, k ) ) return curr;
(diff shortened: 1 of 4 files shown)


vigna:
Methods like those exist in ConcurrentSet. Did you consider implementing the interface instead? 

On August 7, 2015 5:07:13 PM GMT+02:00, Erich Schubert <notifications@github.com> wrote:
>Works like add(k) followed by get(k), but is cheaper.
>
>This is useful for managing unique objects, e.g.
>
>ObjectHashSet<String> unique;
>key = unique.addOrGet(key);
>
>I have wrapped this code in "#if #keyclass(Object)", because for
>primitives it is equivalent to "add(k); return k;".
>
>For hash sets, the code duplication isn't too bad. But for the
>tree-based sets it is massive, so I didn't do these.
>
>Again, untested code, sorry. Consider this to be a feature request.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/vigna/fastutil/pull/4
>
>-- Commit Summary --
>
>  * Add addOrGet(k) to ObjectHashSets.
>
>-- File Changes --
>
>    M drv/OpenHashBigSet.drv (40)
>    M drv/OpenHashSet.drv (54)
>
>-- Patch Links --
>
>https://github.com/vigna/fastutil/pull/4.patch
>https://github.com/vigna/fastutil/pull/4.diff
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/vigna/fastutil/pull/4

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

kno10:
Which ConcurrentSet?
This is not a "compareAndSet" operation. It is not meant for concurrency, but to avoid duplicate copies e.g. of Strings (yes, I know of String.intern) by keeping a Set of previously seen objects. Doing this with a Map<K,K> is possible, but more needs twice as much memory.

kno10:
I have rebased the patch onto the latest. Please see the (tiny) patch to understand why this is not a question of concurrency; but for the purpose of having only one copy of e.g. each string in memory.
It is a combined version of

    public K addOrGet(K k) { if (!add(k)) k = get(k); return k; }

that does `add` and `get` in one lookup.

vigna:
Technically it's two lookups (I mean, in the table), not one. I would rather make a copy of add() and modify it so that in a single lookup it either returns an existing object or it returns the previous version. If you don't feel like it I can do it.

kno10:
The patch already does that.

vigna:
I'm sorry. I thought the change was the line in the comment. I'm new to this stuff :).

One thing. It seems like you modified add() *before* the fix to custom-hash sets. For example, your patch uses KEY_IS_NULL() instead of KEY_EQUALS_NULL() in the the initial check. Please have a look at the modifications to add()/get() and update the patch.

You should also check in some unit test in ObjectOpenHash[Big]SetTest.java.

kno10:
I have updated the patch, including some basic unit testing code (ensuring that addOrGet returns the existing instance for de-duplication). I didn't see the KEY_EQUALS_NULL change in OpenHashBigSet, only in non-Big. Did I miss it?

vigna:
Well, for some reason I don't remember big sets do not have custom-hash versions. Without custom hashes, KEY_IS_NULL is fine.

PR #9, opened 2015-12-02
Add @Deprecated warnings to methods with better alternatives

kno10:
This patch adds @Deprecated warnings to various methods, where a non-boxed alternative exists.
This is useful for finding such locations in user code to improve the code, e.g. if getIntKey() exists, then getKey() will be deprecated and cause a warning.
Conditionals are in place to not deprecate methods for non-primitives.

### DIFF
--- a/drv/AVLTreeMap.drv
+++ b/drv/AVLTreeMap.drv
@@ -803,6 +803,10 @@ public class AVL_TREE_MAP KEY_VALUE_GENERIC extends ABSTRACT_SORTED_MAP  KEY_VAL
 	 
 
 #if ! #keyclass(Object) || #values(primitive)
+	/** {@inheritDoc}
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
 	public VALUE_GENERIC_CLASS put( final KEY_GENERIC_CLASS ok, final VALUE_GENERIC_CLASS ov ) {
 		final VALUE_GENERIC_TYPE oldValue = put( KEY_CLASS2TYPE(ok), VALUE_CLASS2TYPE(ov) );
 		return modified ? OBJECT_DEFAULT_RETURN_VALUE : VALUE2OBJ( oldValue );
@@ -812,6 +816,10 @@ public class AVL_TREE_MAP KEY_VALUE_GENERIC extends ABSTRACT_SORTED_MAP  KEY_VAL
 
 
 #if ! #keyclass(Object) || #values(primitive)
+	/** {@inheritDoc}
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
 	public VALUE_GENERIC_CLASS remove( final Object ok ) {
 		final VALUE_GENERIC_TYPE oldValue = REMOVE_VALUE( KEY_OBJ2TYPE( ok ) );
 		return modified ? VALUE2OBJ( oldValue ) : OBJECT_DEFAULT_RETURN_VALUE;
@@ -1007,6 +1015,11 @@ public class AVL_TREE_MAP KEY_VALUE_GENERIC extends ABSTRACT_SORTED_MAP  KEY_VAL
 			return prev;
 		}
 
+#if #keys(primitive)
+		/** {@inheritDoc}
+		 * @deprecated Please use the corresponding type-specific method instead. */
+		@Deprecated
+#endif
 		public KEY_GENERIC_CLASS getKey() {
 			return KEY2OBJ(key);
 		}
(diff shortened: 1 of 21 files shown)


vigna:
It's maybe a bit heavy-handed, but I think it's a good idea. Two changes are necessary I think:

1) In line with other deprecations, the comment should be "Please use the corresponding type-specific method instead." (Capitalized, with period at the end.)
2) You should be sure that whenever you insert a @deprecated, there is a corresponding comment in the Javadoc, or people will be baffled. For a few methods where documentation is missing I think this can be obtained by adding something like

    /** {@inheritDoc}
     *  @deprecated Please use the corresponding type-specific method instead.
     */

kno10:
I tried to produce better comments (in particular: linking to the recommended replacement instead - getIntKey() etc.) but the preprocessor apparently does not process the comments. That would be the most helpful version of the annotation, "Please use {@see #getIntKey()} instead."

vigna:
No, that unfortunately can't be done. Unless we switch to a completely different macro processor, there's unfortunately no way to have the processor process comments. So comments must be generic (like, "use the type-specific method").

What I meant is that there are a few methods (like put() in maps) that you @deprecated that have no comments. There you need to try to have something with {@inheritDoc}. The rationale is that if the user looks at the Javadoc it should understand immediately why the method is deprecated. I think it's just a few cases.

kno10:
I have updated the patch with {@inheritDoc} and the comment.

vigna:
Happily merged! Thanks for your contribution.

vigna:
I'm seeing that a few get() methods around didn't get the deprecation. Some (like in open hash maps) simply miss the Javadoc. Some (like in AVLTreeMap) were deprecated upstream, but I'm not sure this will cause the deprecation to be reported by the IDE. I think it is necessary to inherit the documentation at a lower level (but I might be wrong).

Moreover, do you know if it is possible to have the deprecation at the *end* of the @inheritDoc-generated documentation, rather than at the start? It's a bit ugly.

kno10:
I tried to find all inherited methods and mark them; but I may have missed some.
I'm not sure if @Deprecated is inherited (I believe it only causes a warning in the overriding method; which may be a feature to allow un-deprecating a method for subtypes maybe?). With {@inheritDoc}, I mostly remember that it didn't work as I wanted, so I ended up not using it in my projects. The placement of @annotations in JavaDoc may be hardcoded - too many things are hardcoded in JavaDoc, unfortunately.

vigna:
OK. The problem is that you exaggerated:

    /** {@inheritDoc}
     *
     * Delegates to a more generic method.
     *
     * @deprecated Please use the corresponding type-specific method instead. */
    @Deprecated
    public boolean addAll( final long index, final COLLECTION c ) {
        return addAll( index, (Collection<? extends KEY_CLASS>)c );
    }

What does this mean? This is the MOST SPECIFIC METHOD. It delegates to a MORE GENERIC METHOD. That's definitely THE method you shouldn't deprecate.

vigna:
Please comment.

kno10:
Sorry, I have to take a longer break.

AFAICT, that deprecation should be removed again. It has the type-specific signature, but calls the less efficient generic version internally (so it may still be good to avoid this) and I assume that subtypes will often override this with a more efficient implementation. I'm not sure if @Override gets inherited (and such an override will likely not have this drawback) and I thus believe it should be removed.
I hope I did not add any others that shouldn't be there - I tried my best only to add it to relevant places only, but I don't know this source code in-and-out.

vigna:
Also here: GET_VALUE and REMOVE_VALUE *are* the type-specific calls. 

Can you check that you did not deprecated any GET_VALUE/REMOVE_VALUE around and fix those deprecated? 

I'm checking every deprecation, but it's taking a lot of time and I'm starting to think I've been too quick to merge this patch.

if #keys(reference) || #values(reference)
#if #values(primitive)
        /** {@inheritDoc}
         * @deprecated Please use the corresponding type-specific method instead. */
        @Deprecated
        @Override
#endif
        public VALUE_GENERIC_TYPE REMOVE_VALUE( final Object k ) { throw new UnsupportedOperationException(); }
#if #values(primitive)
        /** {@inheritDoc}
         * @deprecated Please use the corresponding type-specific method instead. */
        @Deprecated
        @Override
#endif
        public VALUE_GENERIC_TYPE GET_VALUE( final Object k ) { return function.GET_VALUE( k ); }
#endif

vigna:
OK, I completed the review and everything seems in its right place now. I'll remove or add deprecation following user reports.

kno10:
For these two, I added the @Deprecated annotation because they use the non-primitive argument.
So `map.removeInt(1)` will not be deprecated, but `map.removeInt(Integer.valueOf(1))` will.

But the given reason is not appropriate, and this reason for deprecating also is different, and maybe too strict; so feel free to revert these two hunks, too.

vigna:
You haven't understood the point: if keys are objects,  `map.removeInt(Integer.valueOf(1))` should *not* be deprecated. If values are primitive, they are not references, so we entered the outer `#ifdef` because keys are references.

I also discovered that in a number of circumstances you included in the `#ifdef` isolating the deprecation part the `@Override` annotation. This is wrong—the method overrides whether it is deprecated or not.

kno10:
It should have been `#if #keys(primitive)` then, sorry.
Thanks for reviewing.

PR #36, opened 2016-03-23
FJQuickSort needs to use the comparator to choose the partition.

rossjudson:
Was missing the comp parameter. Added a test case for this.


vigna:
Thanks for the fix!

PR #55, opened 2017-01-30
Immutable List and Set

yasserg:
This patch implements immutable versions of List and Set. fastutil
already had helper methods like "IntLists.unmodifiable(list1)" but
the resturned list in this case was just an unmodifiable view of
list1 and not truely immutable. Updating list1 would still result
in changes in the unmodifiable view.

The ImmutableList and ImmutableSet implementations here use Builder
pattern or static methods to collect the data in the collections and
then build the immutable collection from that. Hence in addition to
immutability, they also use the minimum memory required for the
provided size of data.

### DIFF
--- a/drv/ImmutableList.drv
+++ b/drv/ImmutableList.drv
@@ -1,0 +1,297 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package PACKAGE;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Spliterator;
+import java.util.Spliterators;
+
+/**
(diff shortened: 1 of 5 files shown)


yasserg:
@vigna please let me know if you are willing to merge this patch. I already have unit tests and if you're ok with this addition I will add them to patch.

vigna:
What is the advantage with respect to, like,  IntSets.immutableSet(new IntOpenHashSet(set)) ? In particular, I see little incentive to duplicate the hash-table logic (yours is different, and unlikely to be faster).

yasserg:
I have been very busy the past couple of days. Will try to get back on this soon. I think I can prepare a revision that reuses most of the logic in IntOpenHashSet.

vigna:
But, once again, what's wrong with  IntSets.immutableSet(new IntOpenHashSet(set)) ?

yasserg:
We maintain millions of small lists and sets in memory. Using "IntSets.immutableSet(new IntOpenHashSet(set))" means that for each of those we will have a redundant wrapper object which is just overhead.

vigna:
Seems a very narrow case. Possibly you should extrapolate code from OpenHashSet and eliminate all the mutation parts.

yasserg:
@vigna, please take another look. I removed more of the new codes and reused some of your utilities from HashCommon and Hash. After this, the amount of non-boilerplate code is only about 80 lines of code. I think it makes more sense to keep this separate from OpenHashSet, because:
 - ImmutableIntSet determines the minimum table size during construction. OpenHashSet doesn't have this logic and it doesn't make sense for it.
 - RegularImmutableSet (an implementation of ImmutableIntSet) is almost identical to OpenHashSet if we remove all mutation logic from OpenHashSet.  However, given that it is immutable it caches hashcode and uses that for faster hashcode/equals and it has one less field (instead of max fill and load factor it has cached hashcode). We could try to reuse some code here but I don't see net benefit given the class is very simple and there are subtle differences.

PR #68, opened 2017-06-07
Java 8 adaptions

### LINKED ISSUE #64: Java 8 support
Hi,

as you said in #43:
> Well, Java 7 hit the end of life. So Java 8 support it's a matter of survival.

Do you plan to migrate to Java 8? I strongly urge you to - I personally need most of these changes :) I identified several "todo"-items for Java 8, which I would be willing to implement:

- Support of the new Map methods like `putIfAbsent`, `computIfAbsent` etc. (#57)
- Synchronization support of the newly introduced map methods (#48)
- Support of new Iterator interfaces (#43)
- Support of new Function interfaces, e.g. `Int2DoubleMap` should implement Java 8's `IntToDoubleFunction`.

I imagine that one could implement all this in a way which allows you to specify the target Java version. But, while it should be possible, I advice against it; as you pointed out, Java 7 is not the go-to version anymore and future releases should target Java 8.

incaseoftrouble:
Closes #64


drv/Function.drv
```diff
@@ -104,9 +159,13 @@ public interface FUNCTION KEY_VALUE_GENERIC extends Function<KEY_GENERIC_CLASS,
 
 	/** {@inheritDoc}
 	 * @deprecated Please use the corresponding type-specific method instead. */
+
 	@Deprecated
 	@Override
-	boolean containsKey(Object key);
+	default boolean containsKey(Object key) {
```
vigna: This is different from the original methods, which just checks for null (returning null).

src/it/unimi/dsi/fastutil/SafeMath.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+public final class SafeMath {
+	private SafeMath() {}
+
+	public static char safeIntToChar(int value) {
+		if (value < (int) Character.MIN_VALUE || (int) Character.MAX_VALUE < value) {
+			throw new IllegalArgumentException(value + " too big for char");
+		}
```
vigna: All these "too big" are actually "out of range".

src/it/unimi/dsi/fastutil/SafeMath.java
```diff
@@ -0,0 +1,66 @@
...
+			return Float.NaN;
+		}
+		if (Double.isInfinite(value)) {
+			return value < 0.0d ? Float.NEGATIVE_INFINITY : Float.POSITIVE_INFINITY;
+		}
+		if (value < (double) Float.MIN_VALUE || (double) Float.MAX_VALUE < value) {
+			throw new IllegalArgumentException(value + " too big for float");
+		}
+		float floatValue = (float) value;
+		if (((double) floatValue) == value) {
```
vigna: Unless I'm really lost, this == should be !=.

drv/Map.drv
```diff
@@ -142,7 +210,459 @@ public interface MAP KEY_VALUE_GENERIC extends FUNCTION KEY_VALUE_GENERIC, Map<K
...
+
+	default VALUE_GENERIC_TYPE GET_VALUE_OR_DEFAULT(KEY_TYPE key, VALUE_GENERIC_TYPE defaultValue) {
+#if VALUES_PRIMITIVE
+		return containsKey(key) ? GET_VALUE(key) : defaultValue;
+#else
+		V v;
+		return (v = get(key)) != null || containsKey(key) ? v : defaultValue;
+#endif
+	}
+
```
vigna: I think that the logic in the following methods does not suit to fastutil. In fastutil, the default return value replaces for all purposes null, even in object-based methods. This has a number of useful consequences: if the user, for instance, ignores it, he/she gets back the JDK behaviour. But I could create a single marker object as default value and then I could use null as a return value without any confusion between existing mappings and mappings to null.

Thus, specification and implementation of the following methods should be uniform: same semantics of the JDK were "null" is replaced by the default value (as it happens currently for primitive types).

drv/Map.drv
```diff
@@ -142,7 +210,459 @@ public interface MAP KEY_VALUE_GENERIC extends FUNCTION KEY_VALUE_GENERIC, Map<K
 
 	@Deprecated
 	@Override
-	boolean containsValue(Object value);
+	default boolean containsValue(Object ov) {
```
vigna: Similarly to containsKey(), this method should just check for null and return null, not check the class (as the previous implementation in the abstract class was doing).

drv/Maps.drv
```diff
@@ -91,23 +98,12 @@ public class MAPS {
 #endif
 
 		@Override
-		public void putAll(final Map<? extends KEY_GENERIC_CLASS, ? extends VALUE_GENERIC_CLASS> m) { throw new UnsupportedOperationException(); }
+		public void putAll(Map<? extends KEY_GENERIC_CLASS, ? extends VALUE_GENERIC_CLASS> m) { throw new UnsupportedOperationException(); }
```
vigna: A number of "final" qualifiers throughout the code have been removed. They shouldn't. Actually, I prefer *all* arguments that are not modified to be marked as final, as this facilitates the compiler work and makes it easier to follow the code. Please revert the deletions and, whenever possible, use final.

drv/Functions.drv
```diff
@@ -230,7 +250,11 @@ public class FUNCTIONS {
 		public int hashCode() { synchronized(sync) { return function.hashCode(); } }
 
 		@Override
-		public boolean equals(Object o) { if (o == this) return true; synchronized(sync) { return function.equals(o); } }
+		public boolean equals(Object o) {
+			if (o == this) return true;
+			if (o instanceof SynchronizedFunction) synchronized (sync) { return function.equals(((SynchronizedFunction) o).function); }
```
vigna: VERY DANGEROUS. Accesses a synchronized map without synchronization! 

In this cases, a look at the JDK code is a good idea...

drv/Maps.drv
```diff
@@ -324,11 +317,112 @@ public class MAPS {
 		public int hashCode() { synchronized(sync) { return map.hashCode(); } }
 
 		@Override
-		public boolean equals(final Object o) { if (o == this) return true; synchronized(sync) { return map.equals(o); } }
+		public boolean equals(final Object o) {
+			if (o == this) return true;
+			if (o instanceof SynchronizedMap) synchronized (sync) { return map.equals(((SynchronizedMap) o).map); }
```
vigna: VERY DANGEROUS. Accesses a synchronized map without synchronization! 

In this cases, a look at the JDK code is a good idea...

drv/Maps.drv
```diff
@@ -403,6 +494,110 @@ public class MAPS {
...
+		@Override
+		public boolean isEmpty() { return map.isEmpty(); }
+
+		@Override
+		public int hashCode() { return map.hashCode(); }
+
+		@Override
+		public boolean equals(final Object o) {
+			if (o == this) return true;
+			if (o instanceof UnmodifiableMap) return map.equals(((UnmodifiableMap) o).map);
```
vigna: Not dangerous, but frankly unnecessary and possibly dangerous (e.g., the unmodifiable wrapper could do things you wanna do when you access the map).

src/it/unimi/dsi/fastutil/Function.java
```diff
@@ -40,22 +39,36 @@
  * @see java.util.Map
  */
 
-public interface Function<K,V> {
+public interface Function<K,V> extends java.util.function.Function<K,V> {
+
```
vigna: I think we should {@inheritDoc} here, and in the following comment explain what the default method does.

drv/Map.drv
incaseoftrouble: So you say that e.g. `putIfAbsent` should be modified to
```
VALUE_GENERIC_TYPE v = get(key);
if (v != null) return v;
put(key, value);
return defaultReturnValue();
```
Correct?

drv/Function.drv
incaseoftrouble: Oh yes, missed this.

This implies that all implementations are "required" to reject `null` keys?

drv/Map.drv
incaseoftrouble: Similar point as above - this means that null keys are not allowed (or at least not respected) by the default implementation. Is this correct?

drv/Maps.drv
incaseoftrouble: Ah I see. Force of habit. The project I am currently working on primarily discourages excessive use of `final`. I will try to "finalize" everything.

drv/Functions.drv
incaseoftrouble: Ah, good spot, yes.

src/it/unimi/dsi/fastutil/Function.java
incaseoftrouble: Not sure if I entirely follow you here. You mean the `apply` default method? I think its documentation is fine as is. Maybe rewrite it to something like "returns the value associated with this key or null if no value is present, i.e. returns get(key)". The added default methods should have appropriate documentation.

drv/Function.drv
vigna: So: the rationale for type-specific methods is that null is not part of the map (sorry, above should be "returning false"), but it is handled gracefully. Anything else get a OBJ2TYPE() and then it is handled as a primitive type.

Just check the previous code. Those methods were already implemented in the abstract classes, but you probably rewrote them instead of moving them.

drv/Map.drv
vigna: Sorry, the contrary. Null keys are allowed for primitive keys, and they are considered not part of the map. I repeat: please do not modify the previous implementation.

drv/Maps.drv
vigna: Oh well I guess there might be a billion different opinions. Do not force yourself, it's OK, but please do not delete my finals :).

src/it/unimi/dsi/fastutil/Function.java
vigna: Yes, but this method is inherited from another interface. It is not defined in this interface. In this interface we simply define the default version. So we must {@inheritDoc} the specification, and add the documentation about the implementation.

drv/Function.drv
incaseoftrouble: Yes, for primitives its clear. I just meant that in the case of, say, `Object2IntMap` would a `put(null, 1)` be something you want allow to happen? Then the `if (key == null) return false` check should actually only be happening for primitive keys and not all. In the abstract implementation, this was not the case (i.e. `null` keys were not considered by it). I am just not sure if this is intended. If I didn't miss it this is not explicitly documented.

drv/Map.drv
incaseoftrouble: Yes, for primitive values! But (as above) consider a `Int2ObjectMap`. Do you want to allow a `put(1, null)`? Then a `if (ov == null) return false` short-circuit is not correct.

drv/Maps.drv
incaseoftrouble: Definitely. I still am a supporter of `final`ing everything so I just did that ;)

drv/Map.drv
vigna: Difficult question. The point is: why do we have a different code than the primitive case? For me, I would use exactly the same code.

There are some subtle semantics points here. fastutil maps always have settable default return values, which makes the logic "if it returns null or it is mapped to null" logic a bit weird. I mean, why don't we check v != defaultReturnValue()?

So one proposal could be:

final VALUE_GENERIC_TYPE v = get(key), drv = defaultReturnValue();
if (v != drv) return v;
put(key, value);
return drv;

This is possibility that preserves the original semantics in case defaultReturnValue() == null.

The problem is that now the semantics diverges from the primitive case. We have to decide if this is an asset or a liability.

drv/Function.drv
vigna: No, the containsKey() implementation is just for primitive keys.

#if KEYS_PRIMITIVE
    public boolean containsKey( final Object ok ) {
        if ( ok == null ) return false;
        return containsKey( KEY_OBJ2TYPE( ok ) );
    }
#endif

I don't know what you mean by the put(null, 1) example. This problem is only for primitive keys. When handling primitive keys, you check for null on reading and act as if it wasn't in the map. For the rest, OBJ2TYPE().

drv/Map.drv
vigna: ...and, once again, this is the original code:

#if VALUES_PRIMITIVE
    public boolean containsValue( Object ov ) {
        if ( ov == null ) return false;
        return containsValue( VALUE_OBJ2TYPE( ov ) );
    }
#endif

drv/Map.drv
incaseoftrouble: I am strongly against something like `if (v != drv) return v;`. In the documentation of Function, it is written that `null` should always be used to signal absence of a key when `get` is called. If we now introduce such a check we probably can create some weird behavior.

The weirdness of `null` is something most people agree on I think, but we should stick with it.

drv/Function.drv
incaseoftrouble: > No, the containsKey() implementation is just for primitive keys.

D'uh. My bad. I forgot this. Totally agree then.

drv/Map.drv
incaseoftrouble: See above. You are right, I missed that this is only added for primitive values.

drv/Map.drv
vigna: We're not writing in Function—we're writing in type-specific map. Look at the documentation of get() in drv/Function.drv.

incaseoftrouble: Hm. The type specific one returns the default return value but the "standard" one returns null. This discrepancy is weird but okay, I guess, but then it should be consistent, i.e. have the "standard" `putIfAbsent` consider null and the other consider the default return value. Let me think about this for a while, maybe I can come up with a reason for or against it.

One more point:
```
m.put(1, 1);
assert m.size() == 1 && m.contains(1);
m.defaultReturnValue(1);
assert m.size() == 1 && !m.contains(1);
```
This is weird, right?

vigna: Totally. I'm reckoning myself. It is difficult to do these things right (at least, to get a minimum surprise).

vigna:
> On 7 Jun 2017, at 20:09, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> One more point:
> 
> m.put(1, 1);
> assert m.size() == 1 && m.contains(1);
> m.defaultReturnValue(1);
> assert m.size() == 1 && !m.contains(1);
> 
> This is weird, right?

Well, it wouldn't compile, so, yes, it's somewhat weird :). Even assuming you meant m.containsKey(), I don't understand the point.

Maybe you should specify the type of m?

Ciao,

					seba

incaseoftrouble:
I was typing without thinking too much there. I tried to come up with a "counterexample" when we use the default return value in a particular way but it was unfinished. I accidentally pressed submit there. Please forget about this piece of code :)

incaseoftrouble:
I'll just pull the discussion about the role of the default return value in new default-able methods here. New default methods are `getOrDefault`, `putIfAbsent`, binary `remove`, `replace`, `computeIfAbsent`, `computeIfPresent`, `compute` and `merge`. All of those methods have to be considered separately for primitive and non-primitive values. 

Primitive values first (since they are simpler):
 * `getOrDefault`, `putIfAbsent`, binary `remove`, `replace`: clear (we *only* consider `containsKey`)
 * `computeIfAbsent`: This one actually is quite tricky even in this case (for a different reason). Currently, if there is a primitive Java function, this function is used as mapping function. This means that the function can't signal "I don't want to add this key" by returning `null`. We could change this by using functions always mapping to non-primitive values. Putting a `if (function.apply(key) == drv) return drv` here doesn't feel right.
 * `computeIfPresent`: Also clear. If the key is not present (determined by `containsKey`), we just return the drv.
 * `compute`: Tricky again! DRV occurs in two places here. The standard `compute` takes a key and a binary function and, if a value is present, updates the value with `function(key, oldValue)`, and otherwise with `function(key, null)`. I this case, I think using null (for primitive values) is fine, since this is the only way of "telling" the function that there actually is no mapping. If the function returns `null` or there was no value present, the method returns the DRV.
 * `merge`: Quite clear. If the key is not present, put the given value. Otherwise put key and present value into the given function and put the result. Return old value / DRV.

Non-primitive values:
 * `getOrDefault`: I suggest something like `if (!containsKey || value == drv) return default`. I think we actually even need to do it this way to be consistent with the type specific `get` (recall that `!containsKey(key)` implies `get(key) == drv`). The JDK default versions are still consistent with the spec and our "separate" realm of type specific functions is also consistent internally (note that we do not override the Java defaults - something I forgot while initially thinking about the problem).
 * All other methods: Same as above. We use both `containsKey` and `drv` to check "membership" in the type-specific implementation.

vigna:
> On 8 Jun 2017, at 08:14, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> 	• getOrDefault, putIfAbsent, binary remove, replace: clear (we only consider containsKey)

Almost: in line with the current implementation, there should be a test for null that in case of reading behaves as if the map does not contain the key. So, with the exception of putIfAbsent there should be a test for null that replicates the behaviour for a missing key (e.g., getOrDefault() returns default).

> 	• computeIfAbsent: This one actually is quite tricky even in this case (for a different reason). Currently, if there is a primitive Java function, this function is used as mapping function. This means that the function can't signal "I don't want to add this key" by returning null. We could change this by using functions always mapping to non-primitive values. Putting a if (function.apply(key) == drv) return drv here doesn't feel right.

No no. Wait. The replacement of null with a default return value is *only for the data structure*. The behavior of functions is untouched. I don't think there's any way to replicate the JDK behaviour with primitive functions. People willing to have the possibility of eliminating the mapping will use class-based functions.

> 	• computeIfPresent: Also clear. If the key is not present (determined by containsKey), we just return the drv.

OK.

> 	• compute: Tricky again! DRV occurs in two places here. The standard compute takes a key and a binary function and, if a value is present, updates the value with function(key, oldValue), and otherwise with function(key, null). I this case, I think using null (for primitive values) is fine, since this is the only way of "telling" the function that there actually is no mapping. If the function returns null or there was no value present, the method returns the DRV.

Yes--once again, I wouldn't change the function behaviour. But if the function is from primitive types I don't see how you can do function(key, null). In general I believe that if we want to have primitive bifunctions (you mean bifunctions, right?) we must drop all behaviours related to null.

> 	• merge: Quite clear. If the key is not present, put the given value. Otherwise put key and present value into the given function and put the result. Return old value / DRV.

OK.

> Non-primitive values:
> 
> 	• getOrDefault: I suggest something like if (!containsKey || value == drv) return default.

I'm lost at the variable names. drv, default, value?

> I think we actually even need to do it this way to be consistent with the type specific get (recall that !containsKey(key) implies get(key) == drv). The JDK default versions are still consistent with the spec and our "separate" realm of type specific functions is also consistent internally (note that we do not override the Java defaults - something I forgot while initially thinking about the problem).

> 	• All other methods: Same as above. We use both containsKey and drv to check "membership" in the type-specific implementation.

Why both? If you use containsKey() you don't need drv.

Ciao,

					seba

incaseoftrouble:
> Almost: in line with the current implementation, there should be a test for null that in case of reading behaves as if the map does not contain the key. So, with the exception of putIfAbsent there should be a test for null that replicates the behaviour for a missing key (e.g., getOrDefault() returns default).

I can't fully follow. Lets consider `getOrDefault`. For primitive values, the implementation is `return containsKey(key) ? GET_VALUE(key) : defaultValue;`. How would you change it?

> No no. Wait. The replacement of null with a default return value is *only for the data structure*. The behavior of functions is untouched. I don't think there's any way to replicate the JDK behaviour with primitive functions. People willing to have the possibility of eliminating the mapping will use class-based functions.

Misunderstanding here. Lets consider the example of `Int2IntMap`. There, we currently have `computeIfAbsent(int key, IntUnaryOperator mappingFunction)`. What I meant is that we want to have a non-boxing mapping function here on the one hand, but on the other hand, this function can't signal "I don't want to add a value for the key". In the Java equivalent (`computeIfAbsent(Integer key, Function<Integer, Integer> mappingFunction)`, the mapping function can return `null` to signal that the key should not be added.

> Yes--once again, I wouldn't change the function behaviour. But if the function is from primitive types I don't see how you can do function(key, null). In general I believe that if we want to have primitive bifunctions (you mean bifunctions, right?) we must drop all behaviours related to null.

Exactly! That is the problem. The `null` given to and returned by the mapping functions has a special role which might be missed if we go fully primitive.

> I'm lost at the variable names. drv, default, value?

Yes, sorry.

> Why both? If you use containsKey() you don't need drv.

Java does something like `value == null && !containsKey(key)`, to avoid calling containsKey if possible. I'm not sure if we should use this pattern.

incaseoftrouble:
I pushed some changes. Most of the issues should be addressed. Please take a look

vigna:
> On 8 Jun 2017, at 10:20, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> Almost: in line with the current implementation, there should be a test for null that in case of reading behaves as if the map does not contain the key. So, with the exception of putIfAbsent there should be a test for null that replicates the behaviour for a missing key (e.g., getOrDefault() returns default).
> 
> I can't fully follow. Lets consider getOrDefault. For primitive values, the implementation is return containsKey(key) ? GET_VALUE(key) : defaultValue;. How would you change it?

Ahhh you're right, sorry. containsKey() does already the check.

> No no. Wait. The replacement of null with a default return value is only for the data structure. The behavior of functions is untouched. I don't think there's any way to replicate the JDK behaviour with primitive functions. People willing to have the possibility of eliminating the mapping will use class-based functions.
> 
> Misunderstanding here. Lets consider the example of Int2IntMap. There, we currently have computeIfAbsent(int key, IntUnaryOperator mappingFunction). What I meant is that we want to have a non-boxing mapping function here on the one hand, but on the other hand, this function can't signal "I don't want to add a value for the key". In the Java equivalent (computeIfAbsent(Integer key, Function<Integer, Integer> mappingFunction), the mapping function can return null to signal that the key should not be added.

No misunderstanding: I agree with you. With primitive functions, you'll lose that piece of behaviour.

> 
> Yes--once again, I wouldn't change the function behaviour. But if the function is from primitive types I don't see how you can do function(key, null). In general I believe that if we want to have primitive bifunctions (you mean bifunctions, right?) we must drop all behaviours related to null.
> 
> Exactly! That is the problem. The null given to and returned by the mapping functions has a special role which might be missed if we go fully primitive.

It will be missed. There's nothing we can do about it.

> 
> I'm lost at the variable names. drv, default, value?
> 
> Yes, sorry.
> 
> Why both? If you use containsKey() you don't need drv.
> 
> Java does something like value == null && !containsKey(key), to avoid calling containsKey if possible. I'm not sure if we should use this pattern.

I see. You mean


But it's fundamental that you do it THE OTHER WAY AROUND and WITH &&:

value = get(key);
if (value == drv && !containsKey(key)) return default;

incaseoftrouble:
> No misunderstanding: I agree with you. With primitive functions, you'll lose that piece of behaviour.

Well, I guess then I misunderstood you :) Anyway, the question now is: Which version do we want? We can do "use primitive keys when possible but always use generic values"? On the other hand, if someone wants "deleting" behavior, he can always use the generic methods. We should not deprecate them, though, since they are not fully superseded by their primitive counterparts.

> But it's fundamental that you do it THE OTHER WAY AROUND and WITH &&

Definitely. It should (at least now) be as you say. Either `value == drv && !containsKey(key)` or `value != drv || containsKey(key)` (depending on the situation, one is the negation of the other). If not then it is a typo.

vigna:
> On 8 Jun 2017, at 14:27, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> No misunderstanding: I agree with you. With primitive functions, you'll lose that piece of behaviour.
> 
> Well, I guess then I misunderstood you :) Anyway, the question now is: Which version do we want? We can do "use primitive keys when possible but always use generic values"? On the other hand, if someone wants "deleting" behavior, he can always use the generic methods. We should not deprecate them, though, since they are not fully superseded by their primitive counterparts.

No, I would use fully primitive functions. And yes, we should not deprecate (sigh) the other version.

vigna:
> On 8 Jun 2017, at 11:11, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> I pushed some changes. Most of the issues should be addressed. Please take a look

I'd say it's 90% OK. I'm noticing that we could use Objects.equals() in a number of places (maybe it gets intrinsified) and I think still more work is needed for the methods from computeIfAbsent() onwards (there are still some nulls around that should be drv IMHO). But it's really getting in shape.

I'll add some comment about this.

incaseoftrouble:
> I'm noticing that we could use Objects.equals() in a number of places (maybe it gets intrinsified)

Best idea would be to put Objects.equals into KEY_EQUALS and VALUE_EQUALS I think. Then it is most consistent (and you can easily revert it if it doesn't work out as it should.

I pushed some more changes (cleaned nulls I overlooked and added an idea on how to deal with the computeIfAbsent issue).

vigna:
> On 8 Jun 2017, at 15:58, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> I'm noticing that we could use Objects.equals() in a number of places (maybe it gets intrinsified)
> 
> Best idea would be to put Objects.equals into KEY_EQUALS and VALUE_EQUALS I think. Then it is most consistent (and you can easily revert it if it doesn't work out as it should.

Yes, that's what I meant :).

incaseoftrouble:
I pushed some more small things:
 * A bit of polishing
 * KEY_EQUALS and VALUE_EQUALS use `java.util.Objects.equals`
 * Fixed a small corner-case of generic `put` and `remove`
 * Cached default return value more

If you are fine with the changes please tell me, then I'll squash them.

vigna:
> On 8 Jun 2017, at 18:51, Tobias Meggendorfer <notifications@github.com> wrote:
> 
> If you are fine with the changes please tell me, then I'll squash them.

It's all fine, you can squash.

Ciao,

					seba

incaseoftrouble:
Done.

PR #78, opened 2017-08-07
Added possibility to only generate int, long and double primitives

incaseoftrouble:


mdekstrand:
Suggestion: if 'mini' versions of Fastutil are going to be possible, would it be feasible to split fastutil into multiple Maven packages (`fastutil-long`, `fastutil-int`, etc.)? That way projects using Maven can take advantage of the splitting as well.

incaseoftrouble:
Multiple comments on that (only my personal opinion):

 * A rather practical issue: Since parts of fastutil are parametrized by key- and value-type, one would have to decide where to put, e.g., `Int2LongMap` etc. moreover, a lot of underlying classes are shared. For example, every Map needs `ObjectSet`, `ObjectIterator`, etc. for the entry set. The BinIO and TextIO classes are built from fragments, which is not easily mappable to such splitting.
 * What do you define as "advantage"? Since you are using maven, you probably aren't concerned with downloading 10 instead of 5 megabytes. Reducing the size of the jar is the only advantage I see there. To minimize the size of the distribution, you can just shade the built jar; this is way more effective than anything else.
 * The intent of this particular "mini" version primarily is to match the decision of the Java Developers as in providing primitive operations only for the "fundamental" primitives `int`, `long` and `double`.

mdekstrand:
Thanks @incaseoftrouble. My basic idea is to be general rather than special-case, but the particular special case is probably sufficient.

> A rather practical issue: Since parts of fastutil are parametrized by key- and value-type, one would have to decide where to put, e.g., Int2LongMap etc. Moreover, a lot of underlying classes are shared. For example, every Map needs ObjectSet, ObjectIterator, etc. for the entry set. The BinIO and TextIO classes are built from fragments, which is not easily mappable to such splitting; also, they always need some Byte things. This would be solvable but I think pretty hard to do properly.

Good point. I don't use the IO classes so I have mostly ignored them.

> What do you define as "advantage"? Since you are using maven, you probably aren't concerned with downloading 10 instead of 5 megabytes. Reducing the size of the jar is the only advantage I see there. To minimize the size of the distribution, you can just shade the built jar; this is way more effective than anything else.

For our software (@LensKit), we make use of 'LensKit and all dependencies' bundles. This includes things as separate JARs (in a valid local Maven repository, actually). Fastutil is more than half of that bundle storage size.

> The intent of this particular "mini" version primarily is to match the decision of the Java Developers as in providing primitive operations only for the "fundamental" primitives int, long and double.
> 
> Bottomline: I am unsure how this could be realized in a maintainable way and at least I personally don't see merits that would justify the effort.

Perfectly reasonable. If the `mini` version is distributed on Maven we may be able to switch to it and obtain the vast majority of the benefit.

incaseoftrouble:
One related thing: I'm currently putting together a script which you can use to build this "mini" version yourself (only useful if you have your one maven or want to put a jar into your repo).

It basically is a wrapper to `jdeps`, what you can essentially do is the following: You can give it your codebase, then it searches for all references to any fastutil classes and writes them into a file. You can then modify this file how you like (e.g. include some more classes) and then pass this file to the script together with a fastutil jar. Then, the script searches for all transitive dependencies of this class list (for example, if you have "IntSet" in your dependencies, it adds "IntCollection" to the list) and then builds a jar containing exactly those classes.

This might not be relevant to you, but if it is I'd be happy to have a beta tester :)

incaseoftrouble:
@mdekstrand Look at https://github.com/vigna/fastutil/tree/dependency-script maybe this might help you.

Issue #89, opened 2017-12-08
Please add Automatic-Module-Name manifest entries


io7m:
Hello!

It'd be great if `fastutil` published `Automatic-Module-Name` entries in the jar manifest for each module. This would assign the modules stable names to be used in (modular) Java 9 projects.

Of course, full modularization would be better, but that'd involve generating a rather enormous `module-info.java` file to export all of the hundreds of packages.

vigna:
Is that compatible with Java 8? Sorry but I'm entirely unfamiliar with the modularization feature.

io7m:
Hi.

Yes, it's completely ignored on JDKs prior to 9. Essentially, it's a way of saying "This is the module name that I'll be using in the future if I ever modularize the code". It allows people to specify stable dependencies on your code from Java 9 projects without you actually having to publish full modules.

io7m:
Thanks for merging that in.

Any chance you could push a new release?

vigna:
Released with 8.2.0.

io7m:
Thanks!

io7m:
Hm, checking today, something seems to have gone wrong somewhere. The `8.2.0` jars don't seem to have the manifest header:

```
$ wget -O fastutil-8.2.0.jar http://search.maven.org/remotecontent?filepath=it/unimi/dsi/fastutil/8.2.0/fastutil-8.2.0.jar
--2018-05-28 21:53:52--  http://search.maven.org/remotecontent?filepath=it/unimi/dsi/fastutil/8.2.0/fastutil-8.2.0.jar
Proxy request sent, awaiting response... 200 OK
Length: 18800399 (18M) [application/java-archive]
Saving to: ‘fastutil-8.2.0.jar’

fastutil-8.2.0.jar                100%[============================================================>]  17.93M  1018KB/s    in 25s     

2018-05-28 21:54:17 (734 KB/s) - ‘fastutil-8.2.0.jar’ saved [18800399/18800399]

$ unzip -p fastutil-8.2.0.jar META-INF/MANIFEST.MF
Manifest-Version: 1.0
Bnd-LastModified: 1527515258573
Bundle-ManifestVersion: 2
Bundle-Name: it.unimi.dsi.fastutil
Bundle-SymbolicName: it.unimi.dsi.fastutil
Bundle-Version: 8.2.0
Created-By: 10.0.1 (Oracle Corporation)
[...]

PR #95, opened 2018-01-31
Add 'Automatic-Module-Name' entry to manifest

kashike:
adds #89

I've left out change to `CHANGES` as you haven't bumped the version from `8.1.1` and I'm not sure what your versioning plan is.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -156,6 +156,9 @@
 	<target name="jar" depends="compile">
 		<jar jarfile="fastutil-${version}.jar">
 			<fileset dir="${build}" excludes="it/unimi/dsi/fastutil/*IndirectDoublePriorityQueue*"/>
+			<manifest>
+				<attribute name="Automatic-Module-Name" value="it.unimi.dsi.fastutil"/>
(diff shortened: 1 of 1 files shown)


kashike:
@vigna any chance this could be looked at soon?

vigna:
OK, I merged, but I don't understand your comment above about leaving out `CHANGES`.

vigna:
BTW, if you wanna be thanked you have to send me your real name ;).

kashike:
Thanks for merging this, @vigna!

My comment about `CHANGES` was that I didn't make any changes to the [CHANGES](https://github.com/vigna/fastutil/blob/master/CHANGES) file as I didn't know how you handle that file.

My real name is `Joshua Popoff`.

vigna:
Unfortunately you added the module to the standard build, but not to the OSGi build that is used when uploading to Maven Central. Could you have a look?

kashike:
Sorry - I assumed entries from the manifest would have propagated. No experience with OSGi.

PR #118, opened 2019-03-29
Optimized retainAll removeIf and forEach for array backed List and Set

shishu94:
Added optimized methods "retainAll", "removeIf" and "forEach" based on the optimized method
"removeAll" for arrays.
Added test to cover those 4 methods in the IntArrayListTest and IntArraySetTest

### DIFF
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -541,36 +541,66 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		return true;
 	}
 
+#ifdef JDK_PRIMITIVE_PREDICATE
 	@Override
-	public boolean removeAll(final COLLECTION c) {
-		final KEY_TYPE[] a = this.a;
-		int j = 0;
-		for(int i = 0; i < size; i++)
-			if (! c.contains(a[i])) a[j++] = a[i];
-#if KEYS_REFERENCE
-		Arrays.fill(a, j, size, null);
-#endif
-		final boolean modified = size != j;
-		size = j;
-		return modified;
+	public boolean removeAll(final COLLECTION c) { 		return removeIf((JDK_PRIMITIVE_PREDICATE) v->!c.contains(v));
 	}
-
-#endif
+	
+	@Override
+	public boolean retainAll(final COLLECTION c) {		return removeIf((JDK_PRIMITIVE_PREDICATE) v->c.contains(v));
+	}
+#endif JDK_PRIMITIVE_PREDICATE
+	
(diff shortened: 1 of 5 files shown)


vigna:
Did you benchmark the two implementations? I have no idea of the overhead of having contains passed by pointer.

shishu94:
Hello, this was a long time ago :)
At that time i micro benched with JMH this code and a version with specialized code for each function and if i remember correctly the performance was the same (I simply preferred this cleaner version).
But if you need some evidences of this, i can write the bench again.

vigna:
No, it's OK. Are the unit tests covering all the code? And I'll need your real name if you wanna be thanked in the CHANGES.

drv/ArrayList.drv
```diff
@@ -862,7 +892,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 			catch (IndexOutOfBoundsException e) { mThrowsOutOfBounds = e; }
 
 			try {
-				t.add(p, KEY2OBJ(T));
+				t.add(p, KEY2OBJ(T)"""""");
```
shishu94: I think my keyboard got mad here just before doing the commit

shishu94:
Yes the tests cover each case of the added methods

vigna:
OK, please confirm you don't want your name in.

shishu94:
You can add my name. It is DOS SANTOS JERONIMO Leonel
Do you need me to push a correction for the comment i did or can you fix it ?

vigna:
Ouch, ReferenceArraySet does not compile :(.

shishu94:
Ok, so i'll try to setup the build today and check what is happening

vigna:
Argh, I think I've been too quick. There are numerous problem with this pull request. Here

 return removeIf((java.util.function.IntPredicate) c::contains);

the method is deprecated and Eclipse warns that the conversion is unlikely. Does this work?

vigna:
In general, i have warnings for boxing/unboxing, deprecation, etc. and I expect the code to be completely clean of those.

vigna:
I fixed part of the problems (unqualified Arrays calls, missing cast for arrays) but I'm a bit wary at this point.

shishu94:
No problem, as i said since this has been made some time ago, if you suspect something, i will check the build and rework the pull request if needed. I don't have my laptop now, as soon as i can i'll do it.

shishu94:
If the problem if only on c::contains my wild guess is that eclipse does not like to transform the method reference as an IntPredicate. Replacing c::contains with "v -> c.contains(v)" should be fine.

Issue #116, opened 2019-03-07
Add an intStream() method on Int collections, longStream() on Long Collections, etc


techsy730:
Java has primitive stream types for int, long, and double, that let you use the operations of stream without boxing/unboxing. It would be nice for the int, long, and double collections to export a stream that natively works in those types.

I think there should be a way to do this without having to re-implement the entire streaming framework implementation, but if it turns out there isn't, feel free to close this as infeasible.

vigna:
Yeah, that had been on the list for a while. Sorry, i'm just overwhelmed😭

techsy730:
I may have some time to tackle this along with the spliterator thing. Do you mind if I try my hand at it and maybe send you a pull request in a week or two?

vigna:
> On 7 Mar 2019, at 10:08, TechSY730 <notifications@github.com> wrote:
> 
> I may have some time to tackle this along with the spliterator thing. Do you mind if I try my hand at it and maybe send you a pull request in a week or two?

Go for it!

Ciao,

					seba

techsy730:
I've made a bunch of progress on this. I have added default Spliterator methods for Collection, List, Set, and SortedSet. I have also added a specific implementation for ArrayList, as will as intStream, longStream, and doubleStream as appropriate to the type-specific interface.
While I was at it, I also added a bunch of new default methods (like sort on the list types that delegate to the type-specific array sort in XArrays). If you are not comfortable with that, I can carve them out.

I'm not ready for a pull request yet. I still want to add Spliterator implementations for, at minumum, ArraySet, ArrayMap, BigList, and BigArrayList. If I can get to the Hash based ones, that would be nice too, but those are a bit trickier. And of course I need to still test the dang things. And I got the version code messed up (I annotated it @since 8.1.0, but I see now we are already on the 8.2s), but just wanted to let you know it is coming along.

https://github.com/techsy730/fastutil/tree/spliterators

Issue #115, opened 2019-03-07
Collection implementations don't override spliterator()


techsy730:
The various Collection (and Collections of Map (keySet(), etc)) implementations don't appear to override spliterator(). This means they will inherit the quite inefficient generic default implementations from List, Set, etc. This makes using stream() operations potentially quite inefficient.
Many of the types can abuse the internals of the collection to work much faster.

I get that the stream() (currently the only real user of Spliterators) on anything other then Object, int, long, and double Collections are going to involve either a boxing or a casting (due to only those 3 primitives being "recognized" by java.util.stream, and Object being Object), but for these four cases it would be nice to have.

Issue #98, opened 2018-04-01
Primitive streams


benmccann:
`DoubleArrayList.stream()` returns a `Stream<Double>` which converts from primitive to Object. It would be nice if it returned a `DoubleStream` which is a stream of primitives. If the existing `stream` method cannot be changed then it would be nice to add a new method which returns the primitive `DoubleStream`

vigna:
Yeah, that would be a good thing. We're introducing Java 8 stuff a bit at a time, when we have time. Would you beta test an implementation in case?

incaseoftrouble:
Note: The primitive streams only exist for the "base" primitive types (`int`, `double`, `long`).

vigna:
That's probably something that would be worth implementing everywhere. Looks like a lot of methods tho.

benmccann:
There are only three primitive stream implementations (`DoubleStream`, `IntStream`, `LongStream`), so I'm imagining there would be three methods. I'm not sure when you say there would be "a lot of methods" if three is a lot or if you were envisioning other methods?

vigna:
> On 2 Apr 2018, at 18:42, Ben McCann <notifications@github.com> wrote:
> 
> There are only three primitive stream implementations (DoubleStream, IntStream, LongStream), so I'm imagining there would be three methods. I'm not sure when you say there would be "a lot of methods" if three is a lot or if you were envisioning other methods?

We have to return a Stream. DoubleStream etc. has a billion methods. Someone will have to write them :).

Ciao,

					seba

benmccann:
Do you have to actually implement `DoubleStream`? You might be able to call `java.util.StreamSupport.doubleStream` to create the `Stream`

vigna:
OMG, I didn't even know that thing existed. Thanks!

lwhite1:
@vigna When you implement this I would be happy to beta test.  

And, thanks for the awesome library.

techsy730:
As I mentioned in a dup bug I filed, https://github.com/vigna/fastutil/issues/116, I am working on this.
It's not done yet, but the basics are implemented.
https://github.com/techsy730/fastutil/tree/spliterators

vigna:
The current commit has streams, so anybody interested please test the implementation—comments are welcome.

PR #120, opened 2019-04-09
Spliterators and primitive streams

techsy730:
Here it is, spliterators and primitive stream methods for collections.
Also fill in some other Java 8 methods.

Overview of the change: https://docs.google.com/document/d/1yMf5LSajEQ4i4odnwWrXDPPl7OmjX1Fj2MpHzhCAB5c/edit?usp=sharing

This change addresses https://github.com/vigna/fastutil/issues/116, https://github.com/vigna/fastutil/issues/115, and https://github.com/vigna/fastutil/issues/98

IntArrayList.intStream().sum() is about 3.5 times faster then java.util.ArrayList<Integer>.stream().mapToInt(Integer::intValue).sum()
(1ns per element for ArrayList<Integer>, .28ns per element for IntArrayList)

If you think the changes are substantial enough to warrant a bump to 9.0.0, may I suggest that we require Java 9 for that so that we can take advantage of its high performance methods for arrays and index checks?

I am not content with how the disambiguation methods (such as forEachRemaining(...fastutil.ints.IntConsumer) delegates to forEachRemaining(java.util.function.IntConsumer)) turned out. But without this, calling forEachRemaining(...fastutil.ints.IntConsumer) would cause a compile-time error as it would be ambiguous between java.util.function.IntConsumer and java.util.function.Consumer<Integer>. If you don't think this disambiguation is worth it (for example, you wanted to deprecate ...fastutil.ints.IntConsumer), then I can remove it and the cruft associated with it.

Apologies for doubling the size of TODO. I do plan on getting to the items I listed in it, but I didn't want to block any longer on this.


szarnekow:
> require Java 9 for that so that we can take advantage of its high performance methods for arrays and index checks

Out of curiosity: Which methods are you referring to?

vigna:
So. This is an *enormous* amount of code that will require time to be reviewed. There's even a class (C) Google (huh?).

I think you need to write a document explaining the design choices and how they were implemented. I cannot navigate blindly through all of this code.

vigna:
And, how much of the code is covered by unit tests?

vigna:
I see Objects.checkIndex() is intrinsified in Java 9 (the other check methods do not—bummer). Thank you for the tip (and for all the code).

simlu:
Wow, this looks terrific! Just did a quick test and for what we're doing the speed difference is significant. Looking forward to having this released (after a thorough review)!

techsy730:
Sure, I'll write up a document explaining the new classes and methods and why I coded them as I did and link it here, since yeah, there is a lot of code to go on with little context as to why it is the way it is.

As for unit test coverage, I made sure that almost all of it was covered by tests. Either by unit tests or the runnable main tests. I am afraid I don't have hard code coverage numbers at the moment, but I can probably get those if you really want hard coverage. But I did make sure to test all Spliterators, forEachRemainings, tryAdvances, and trySplits (by way of their usage through Streams)

techsy730:
Also, if the "Copyright Google" is unnerving, I can change that part of the blurb to be under my name (the email is set to my google.com address, so it is still credited to google). Google isn't that strict on copyright.
Also, no matter what, this will still be under Apache 2.0, same as the rest of the code, so you can still do whatever you want with it.

vigna:
Actually, no, it's not, but I would like to know why that source is there.

techsy730:
Basically, it is because I am contributing to this as a Google employee, and as such the author for copyright purposes is Google. That's just our policy for open source contributions "on company time".

Again, it can either by under "Google LLC" or "C. Sean Young", whichever you would be more comfortable with. The exact wording of the header is not important.

techsy730:
> > require Java 9 for that so that we can take advantage of its high performance methods for arrays and index checks
> 
> Out of curiosity: Which methods are you referring to?

For Array methods

[java.util.Arrays.equals(T[], fromIndex, toIndex, T[], fromIndex, toIndex)](https://docs.oracle.com/javase/9/docs/api/java/util/Arrays.html#equals-java.lang.Object:A-int-int-java.lang.Object:A-int-int-)
and
[java.util.Arrays.compare(T[], fromIndex, toIndex, T[], fromIndex, toIndex)](https://docs.oracle.com/javase/9/docs/api/java/util/Arrays.html#compare-T:A-T:A-)

Of which the primitive versions are vectorized, which is much faster then we can do with "regular" Java code.

For Index checking

[java.util.Objects.checkIndex(index, Length)](https://docs.oracle.com/javase/9/docs/api/java/util/Objects.html#checkIndex-int-int-), checkFromToIndex​(fromIndex, toIndex, length), and checkFromIndexSize​(fromIndex, size, length).

Currently only one of them is intrinsicified (I think the basic checkIndex one), but possibly in the future more of them can be.

techsy730:
A design overview doc is incoming; just putting on the finishing touches to it

vigna:
Well, historically contributors of a significant amount of code have been signing a copyright release form. It would have been a good idea to tell me in advance you couldn't do that.

techsy730:
Oh, I'm sorry. I didn't realize there was a process for that. Let me know what I need to do and I will run it by our open source team.

vigna:
It's not a big issue now. It was with LGPL. But let's see what we can do.

techsy730:
Just FYI, Google doesn't require me to sign any CLA for Apache 2.0 licensed projects. If you do on your end, that is fine, I can go through that process. But I am clear to contribute to Apache 2.0 licensed projects on our side.

techsy730:
Ok, here is the design overview for this change:
https://docs.google.com/a/google.com/document/d/e/2PACX-1vScXXgVe0MBQYxLsxd8csFRhxE6nKF4BHb_9D_NiIKIh0LkMrxN0QVF0Dp6Dbs24M24Deikp1XfoezM/pub

Apologies for the requirement to sign in; not sure how to turn that off.

simlu:
Yeah, getting a perm denied...

techsy730:
Ok, let's try this
https://docs.google.com/document/d/1yMf5LSajEQ4i4odnwWrXDPPl7OmjX1Fj2MpHzhCAB5c/edit?usp=sharing

Feel free to leave comments. I've made notes about design choices I am not content about and am willing to consider alternatives.

techsy730:
I think something I am going to try is on different branches, do those alternatives mentioned in the doc, and see if it makes the API simpler and less hacky.

techsy730:
Please don't merge the change as it; I think I have a way to get rid of the awful ImplementsSpliterator hack

techsy730:
I think I have a way to get rid of the awful ImplementsSpliterator hack, if you want to wait for that.

vigna:
OK. It'll take time to review and merge this in, so take your time.

techsy730:
Ok, here is the change I am proposing to get rid of ImplementsSpliterator, and also makes the interface methods a lot cleaner:
https://github.com/techsy730/fastutil/pull/1/files

If you think this is cleaner, then I will merge this into this branch and thus pull request.

techsy730:
I've gone ahead an merged in the change to remove ImplementsSpliterator

Major changes:
- Iterable.ImplementsSpliterator and all sub-interfaces have been removed
- The default implementations of spliterator and intSpliterator have simplified greatly
- New methods and classes in Spliterators, one wrapping an Iterator as a Spliterator, the other wrapping a Spliterator as a Iterator. These methods are analogous to java.util.Spliterators.spliterator and iterator

The design document has been updated accordingly.

benmccann:
@techsy730 your PR will need to be rebased. I'm not affiliated with the project, but I wonder if it might be possible to split it into multiple PRs to make it easier to review? I'm very thankful for the work you've done here and excited to see it make its way in

vigna:
BTW, I didn't forget you... just swamped. I'll have a window to look at this in July.

techsy730:
@benmccann It is possible, but it would be tricky because of how interconnected the commit history is. Furthermore, the Java Spliterator is pretty pervasive; it is hard to implement one part without having the other parts implemented too.
Perhaps I can do something like
Add primitive spliterator interfaces + codegen changes; add primitive spliterator() methods and utility classes; add primitive stream methods.
But would that make it be any cleaner? Doing it like this will result in PRs with a large amount of unused code until all of them go in.

Something I will absolutely look at doing is to split off the List sort method changes into their own PR; they can stand alone.

techsy730:
Ok, rebased to HEAD, and split off the List.sort and Arrays sorting methods into their own pull commit; https://github.com/vigna/fastutil/pull/130

vigna:
I'd release 8.3.0 as it is now. I'll try to go through these changes, even if I don't know if I can do it before September. At that point we could bump to Java >= 9 and the release to 9.0.0. It's a lot of code.

Which are the high performance methods for arrays and index checks?

I'm trying to remember why IntConsumer exists in the first place. It was a long time ago. LOL.

techsy730:
@vigna The high performance methods in Java 9 I am talking about are
java.util.Objects.checkFromIndexSize, checkFromToIndex, and checkIndex (only one of these is currently intrinsicified, but we should still use all of them in case others get it in the future)
java.util.Arrays.equals (overloads accepting bounds were added in Java 9)
java.util.Arrays.compareTo

vigna:
Ah ok. I was guessing that but since only one was intrinsified I thought I was missing something. Certainly starting to use them is a good strategy.

CHANGES
```diff
@@ -17,6 +32,7 @@
   easily by static imports. The old versions have been deprecated.
 
 - New methods to unwrap iterators into big arrays.
+>>>>>>> mainline/master
```
benmccann: is this a merge conflict?

vigna:
I'll finally get time to put this into the release. I'd stay for Java 8 at least this round. You'll hear from me soon...

vigna:
"Soon" is a month ago but yes, you'll hear from me :).

vigna:
And it took one year. Sorry, work, COVID-19, and the rest.

This PR is the largest contribution ever entering fastutil. I've read your support documentation and worked a bit through the code. The contribution is clearly of excellent quality and introduces a much-needed set of features to fastutil.

In terms of code style, the only differences I can see are for-loops with ++i instead of i++ and braces around one-line body of conditionals, etc. Not a big issue, I can do a pass by myself to fix those.

Two files have a Copyright by Google. In general, I've always asked contributors to fastutil to release the copyright to me, GNU style. Indeed 90% of the code you added is in existing files with my copyright, so I think it should not be a big problem. Do you need to ask your employer maybe? 

I invite you to put an @author tag in all files in which you contributed code.

Finally, there is a dire need for a section in overview.html that explains the basics of the design (as you did in the document above). Would you be so kind to write it? Even a rough draft is OK.

And of course, in one year a few conflicts arose. But those should be easy to fix (I hope).

vigna:
Oh, a last question: how much coverage of the new code did you get with your unit tests?

techsy730:
Thanks for taking a look. :)
I am busy with work right now, but I should be able to address your comments during the weekend.

vigna:
Well take your time. I took 18 months... 🙈

techsy730:
Ok, I fixed the merge conflicts, updated the copyright headers, added a section to overview.html about the spliterators and stream methods.

Getting code coverage has been a bit tricky as the tests are spread across the unit tests and the legacy static main method tests (which I think I have a way to integrate into the unit tests, I will make a new PR with that).
I'll post another note with coverage numbers I can get.

techsy730:
Oh, something I would like to add before this being accepted are toList and toSet methods on IntArrayList, IntArraySet, and IntOpenHashSet (and their corresponding Long and Double versions), that take a primitive stream and terminates it collecting it into a list/set. (It won't return a Collector as there is no primitive Collector equivalent in the Java API, it would just have to take the stream)

vigna:
That's amazing!

techsy730:
Ok that is ArrayList.toList() and OpenHashSet.toSet() collector-ish methods are done (and true Collectors for the object based versions)

However, I'm on the fence about ArraySet:
First, I need to decide whether to deduplicate or not.
If I do, I have three options:
1. Call IntStream.distinct, which currently boxes and unboxes into a object based map, which goes against the "ethos" of fastutils
2. First accumulate into OpenHashSet and then copy into ArraySet after it is done, which preserves linear time performance and no boxing, but does have an extra copy step at the end
3. Accumulate into ArraySet directly and take a quadratic loop (linear time search every insert).

If I were to not deduplicate, it would be easy to create invalid state ArraySets.
While asking for uniqueness from the constructors is easy enough to ask callers to verify, in Streams it is not because of the arbitrary operations you can do on stream pipelines.

I'm wondering if ArraySet should even get a collector because ArraySet's with lots of elements are a big performance no-no.

vigna:
Er... why not accumulate in an array, radix-sort it, and deduplicate it in a single pass? That's basically linear, and basically with no additional space (you can use any sort, in fact, like, the platform sort).

techsy730:
Good idea with the sorting. That way I just need to ensure that no two adjacent items are equal.
I will still have to dump to a hash set for large Object streams (as not all Objects are Comparable), but for primitives a sort is great.

As for code coverage, most of my "coverage" was in the form of additions to the main method "stress tests". While coverage using those is pretty good (the version of Eclipse I am using doesn't give delta code coverage over new code, so I can't give a hard number there), I think I should also make them proper unit tests (since we are moving that direction anyways).

incaseoftrouble:
sorry for deviating from the original issue a bit: What definitively should also be in there eventually is property based testing. We can simply use the java collections as reference, generate a stream of random mutations, and check for fidelity.

vigna:
Yes, equal behavior with java.util is what the messy main() tests are trying to achieve...

vigna:
I checked a bit around and @implNote etc. are non-standard. They must be enabled with an option in the tool javadoc and might change in the future. I would refrain from using them in the Javadoc documentation—just add a header "Implementation note" if you need it.

techsy730:
While @implNote and friends are non-standard (and indeed the docket warns you as such), the ability to make your own Javadocs tags is standard.
https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#options (search for "-tag tagname").

That being said, since the Javadocs tool recognized @implSpec and friends as "special", would you rather rename them to something that isn't recognized? Or would you rather refrain from using this Javadocs feature at all?

vigna:
Ok. So essentially as long as we put in build.xml all the relevant -tag options those tags will always work (even if the jdk stops using them). So let's go on. Can you modify the javadoc target of build.xml so they're processed correctly?

techsy730:
Looks like I already did update `build.xml`

And I will get on the spliterator and stream unit tests this weekend, as well as clean up the ArraySet collector methods.

techsy730:
I've got the merge conflict worked out and have a few more tests for the unit test classes, and then I should be ready for review.

techsy730:
Ok, for some code coverage numbers:

- `IntArrayList.Spliterator` has 71.3% coverage. The only thing untested was the `tryAdvance`, as the stream framework usually goes through `forEachRemaining`)
- `IntOpenHashSet.Spliterator` has 51.0% coverage. The only things untested were `skip` (it would be ugly to test this on a spliterator without an ordering assurance) and `tryAdvance` (same reason as above)
- `IntSpliterators` has 51.4% coverage (Most of the uncovered parts are the widening spliterators, like ShortSpliterator -> IntSpliterator wrapper, or the abstract classes)

techsy730:
On second thought, I should add the spliterator tests to the short or byte classes. That way the widening logic to ints can be covered.

(Unfortunately my dev machine just died. I have a spare but setting things up on it will take time)

vigna:
Hey, how are things going with the new machine? :)

techsy730:
I'm more or less comfortable with the state as is being reviewed. Of course I need to fix that merge conflict too.

Machine still in the shop. Haven't had time to setup my other machine with this project yet. I can do that this weekend.

techsy730:
Ok, merge conflicts fixed and Java 8 methods added to ImmutableList.

vigna:
Line 146 of BigList.drv seems a bug to me. You should use length(a), not a.length—the latter is just the number of segments. Right?

vigna:
I merged, but could you add a couple of examples of usage at the end of "an example" section at the end of the overview? I think that would really be helpful.

PR #127, opened 2019-06-21
Add an identity-based hash strategy

kashike:
This introduces a new `Hash.Strategy` that is identity based.

### DIFF
--- a/src/it/unimi/dsi/fastutil/Hash.java
+++ b/src/it/unimi/dsi/fastutil/Hash.java
@@ -105,6 +105,36 @@ public interface Hash {
 		 * @return true if the two specified objects are equal with respect to this hash strategy.
 		 */
 		boolean equals(K a, K b);
+
+		/**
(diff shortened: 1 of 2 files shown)


vigna:
What are the advantages of an identity hash strategy over Reference containers? Just asking, the code is fine (and if you can email me your real name I'll add it to the CHANGES).

kashike:
A wrapper (I assume this is what you mean by "reference container") could be used, but that is more "boilerplate" and you need to implement `equals`/`hashCode` yourself again.

`equals` and `hashCode` implementations can, at times, be (very) slow when they're using some sort of "builder" to check equality/generate a hash.

liach:
i believe this strategy addition is better than extra wrapper objects memory-wise when the number of elements is large.

kashike:
@vigna Any update to this?

vigna:
No,  by "reference container" I mean something like "Reference2LongOpenHashMap", which use an identity-based hash strategy by default.

PR #134, opened 2019-08-03
Fix any issue where longs were using a smaller radix sort threshold

techsy730:
It was intended that longs and doubles had the greatest radix sort
threshold, but it got swapped with ints' thresholds accidentally.


techsy730:
Not a huge deal; just will be missing out on some savings for sorting int arrays

PR #139, opened 2019-11-12
FIFO Queue Peek Function

Speiger:
Added Peek function for faster read only operations. (Like saving a queue)

This is my first try at this, so I tried to keep it as simple as possible with no new imports.
(Trying to get used to it)

Anyways this is really useful function that is really needed.
This function alone can replace a lot of "LinkedList" usecases (in combo what already exists)

I hope this code is good enough.

### DIFF
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -147,6 +147,18 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		array[--start] = x;
 		if (end == start) expand();
 	}
+	
+	/** Allows read-only access into the queue.
+	 * @param index the index to peek at.
+	 */
+	public KEY_GENERIC_TYPE peek(int index)
+	{
+		if(index >= size()}
+		{
+			throw new NoSuchElementException();
+		}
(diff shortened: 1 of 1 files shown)


drv/ArrayFIFOQueue.drv
```diff
@@ -147,6 +147,18 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		array[--start] = x;
 		if (end == start) expand();
 	}
+	
+	/** Allows read-only access into the queue.
+	 * @param index the index to peek at.
+	 */
+	public KEY_GENERIC_TYPE peek(int index)
+	{
+		if(index >= size()}
```
vigna: I expect unit tests. This won't even compile.

Speiger: Eh could you explain why?
Also technically this is not really needed?

JanStureNielsen: Bracket instead of paren...

Speiger: Yeah ok... Github editing is bad (and me too with it). And I hadn't had the time to actually make a real project out of it. Honestly also because I was scared of making bigger changes since i am new to this type of coding.

PR #143, opened 2020-01-17
Fix SafeMath#safeDoubleToFloat for values <= 0

Spottedleaf:
MIN_VALUE is actually the positive closest-to-zero value that a float can represent. So this function would break on 0 and negative float values.


vigna:
If you can send me your name I'll add a thank to the CHANGES file.

PR #145, opened 2020-03-13
Some QOL features

Speiger:
-Added: moveToFirst & moveToLast function which will put the items if present at the start/end. If not present nothing will happen.
-Added: reverseIterator

### DIFF
--- a/drv/OpenHashSet.drv
+++ b/drv/OpenHashSet.drv
@@ -1018,6 +1018,60 @@ public class OPEN_HASH_SET KEY_GENERIC extends ABSTRACT_SET KEY_GENERIC implemen
 		if (ASSERTS) checkTable();
 		return true;
 	}
+	
+	public void moveToFirst(final KEY_GENERIC_TYPE k) {
+		if(size == 0) {
+			return;
+		}
+		if (KEY_EQUALS_NULL(k)) {
+			if (containsNull) {
+				moveIndexToFirst(n);
+			}
+			return;
+		}
+		final KEY_GENERIC_TYPE key[] = this.key;
+		if(KEY_EQUALS_NOT_NULL(k, key[first]) {
+			return;
+		}
+		// The starting point.
+		int pos = KEY2INTHASH(k) & mask;
+
+		// There's always an unused entry.
+		while(! KEY_IS_NULL(key[pos])) {
+			if (KEY_EQUALS_NOT_NULL(k, key[pos])) {
+				moveIndexToFirst(pos);
+				return;
+			}
+			pos = (pos + 1) & mask;
+		}
+	}
+	
+	public void moveToLast(final KEY_GENERIC_TYPE k) {
+		if(size == 0) {
+			return;
+		}
+		if (KEY_EQUALS_NULL(k)) {
+			if (containsNull) {
+				moveIndexToLast(n);
+			}
+			return;
+		}
+		final KEY_GENERIC_TYPE key[] = this.key;
+		if(KEY_EQUALS_NOT_NULL(k, key[first]) {
+			return;
+		}
+		// The starting point.
+		int pos = KEY2INTHASH(k) & mask;
+
+		// There's always an unused entry.
+		while(! KEY_IS_NULL(key[pos])) {
+			if (KEY_EQUALS_NOT_NULL(k, key[pos])) {
+				moveIndexToLast(pos);
+				return;
+			}
+			pos = (pos + 1) & mask;
+		}
+	}
 
 #endif
 
@@ -1350,6 +1404,15 @@ public class OPEN_HASH_SET KEY_GENERIC extends ABSTRACT_SET KEY_GENERIC implemen
 	@Override
 	public KEY_LIST_ITERATOR KEY_GENERIC iterator() {
 		return new SetIterator();
+	}
+	
+	/** Returns a type-specific list iterator on the elements in this set, starting from the last element.
+	 * Please see the class documentation for implementation details.
+	 *
+	 * @return a type-specific list iterator starting at the last element.
+	 */
+	public KEY_LIST_ITERATOR KEY_GENERIC descendingIterator() {
+		return new SetIterator(key[last]);
 	}
 
 #else


Speiger:
@vigna please some opinion on this

vigna:
I'm very busy, sorry. There's a big pull request open since six months. You can reopen the other one of yours is you want.

1) Your descending iterator is not a descending iterator. Starting from the last element does not give you a descending iterator—it simply gives you an ascending iterator starting from the last element.

2) There's a key[first] that should be a key[last]—clearly you didn't write unit tests for this, and  it is not a good practice. 

3) I don't think it is a good idea to have move methods who don't tell you whether the element was or wasn't in the map.

Speiger:
Yeah I can understand that. Though a "I am busy but saw it" would have been nice. Feels less personal especially if you answer to issues. Though I won't open it again for the reason yes its a good idea but that should have been done a bit more extensivly with these kind of things.

1: Yep, I am horrible with naming stuff like that. TopIterator would be my other guess. So I looked at the LinkedList and that was the name that made at least more sense since you will use it to decend. And it was a afterthought.
2: 🤦‍♂ Yes I did this on whimp in github outside of this tiny bug the logic is right.
3: Honestly no, they have the same reason to exist like the "addTo" functions in Maps that have a number in the value side. It just saves performance to constantly check if the element is present.
I am right now using the LinkedHashMap as a priorityqueue for elements to be rendered and clicked on. Sadly javas LinkedHashMap has either any access or nothing aas option. What I wanted to add is reduce overhead and increase QOL. Thats the only issue with this library its powerful but if you want to use one of the QOL functions you are kinda forced to make incompats to other things or have annoying code or break good coding pratices. (Object2IntMap fields if you want to use the addTo function automatically turn into Object2IntOpenHashMap fields. And this shouldn't be a thing ofc this commit does not help it)

About unit tests. I am realtivly new to them, but honestly to improve this library I should maybe get into these. Anyway the issue will be fixed in a second Thanks

vigna:
OK. Why iterator(lastInt()) is bad? I don't see why adding a method to do something like that.

3: I don't understand. There are clearly different exit point depending on whether you find the element or not.

Speiger:
```java
for(ListIterator<Object> iter =set.iterator(set.lastKey());iterator.hasPreviouse();){}
```
Honestly it was a afterthought of why not add this in there since its very likely to be used.

3: Well maps actually do not have this issue since you can say getAndMoveToFirst. So if the key is present then you can move it to first at will.
But with sets its like this:
```java
if(set.contains(object))
{
    set.addAndMoveToFirst(object); //<< Because I dont want to add it if it is missing. Because if it is not present then it shouldnt be moved to first place. Because bugs
}
```
I use this as a "focus order" set. List could be used to but removing stuff from a list is slower.
But it feels like this:
```java
if(!map.containsKey(object))
{
    map.put(object, new ArrayList<Object>());
}
List<Object> list = map.get(object);
```
Even so this is the worst form of lazy loading i know.

I dunno if I get my point across. But this example feels like that.

incaseoftrouble:
Some comments:

1) Relying / modifying iteration order on sets (which, by definition, do not have an intrisic notion of ordering) seems dangerous. Are you sure you don't actually want to use lists (java's reworked array lists have O(1) armortised removal if I remember correctly), priority queues, or linked hash sets (`if (set.remove(element)) set.add(element);` (ab)using insertion order guarantees)?
2) The new methods don't implement a collection interface, as such it would (often, not always) be bad practise to use them outside of the library (your code should not depend on the particular implementation details of a collection).
3) The whole motivation screams "premature micro-optimization" to me. Checking containment in a set really is _stupid_ fast, so even disregarding above points, `if (set.contains(object)) set.addAndMoveToFirst(object);` will probably not have a noticable impact on your application if you don't call this several million times. I really suggest you to properly profile the differences in your application (you can just build fastutil from your current branch and compare execution times).

PS: I do think that methods like `addAndMoveToLast` are not particularly good either, but this is a different topic.

Speiger:
Well my point is: To use features the library offers I have to go ways which are not suggested generally.
And I would like to have them easier to use/access. QOL.
If these features are generally good or bad is a different topic, but if you have to use bad ways to access these extra quality of life features well guess what happens. Bad code in the first place.

Also your remove&add works only in 1 direction. You can only move objects to last so you have to invert iterators etc. Not helping the situation. I am trying to get rid of that problem in the first place.

incaseoftrouble:
> Well my point is: To use features the library offers I have to go ways which are not suggested generally.

I don't see how this applies here. Writing `if (set.contains(object)) set.doSomething(object);` is perfectly recommended. You are (presumably) not writing scientific high performance code. For usual code, readability, maintainability, overall structure is much more important than that last bit of performance.

>  And I would like to have them easier to use/access. QOL.

With previous aspects in mind I personally would argue that this would actually decrease QoL. More methods -> less usability, less maintainability, etc. While this is a small method, its one more thing to take care of. You have witnessed yourself that the maintainer of this repo is quite busy. Plus, again, this method is not exposed in the collection interface so it encourages patterns usually considered bad.

> Also your remove&add works only in 1 direction. You can only move objects to last so you have to invert iterators etc. Not helping the situation. I am trying to get rid of that problem in the first place.

Which is "my" remove? I don't yet know your use case, I was simply suggesting some potentially more suitable datastructures for cases where you want order and uniqueness.


Just to be clear, my main point is: Please try whether `if (set.contains(element))  set.addAndMoveToFirst(object);` is indeed that much slower than `set.moveToFirstIfPresent(object)`. If there is no noticable difference, I don't see an upside to including the new method. Moreover, I'd be genuinely interested in the use case. To me, the function seems rather specific, in which case forking probably is the preferred way.

Speiger:
Eh I am writing my own gameengine and I use this library for the critical points to get the most performance out. This specific issue is mostly for UI, so not "that" critical but I plan to use it in other parts (these specific features) to get faster performance out of it.

So you could say. No I don't do scientific stuff. But performance is a thing I have to watch out for.

And please don't start the language argument or my choice of how I do things.

incaseoftrouble:
> So you could say. No I don't do scientific stuff. But performance is a thing I have to watch out for.

Then you probably should know that premature optimization is bad and you should run proper benchmarks (JMH or real world). I'm genuinely curious how large the impact is. BTW (proper) game engines indeed are science.

> And please don't start the language argument or my choice of how I do things.

I deliberately am not starting this argument. I didn't tell you "xy is bad" but rather "xy is bad practice generally (but there might be exceptions)". This is why I suggested (!) forking: If you are at the stage of developement where you reliably determined that such micro-optimizations matter you probably want to tailor datastructures to your specific use cases anyway. Especially for an involved project as an engine. This is (in my experience) much easier if you fork. I don't care how you want to do what. I'm just suggesting.

However, a library is meant to be used by many. So, whether a change is good depends not only if one person needs the feature but rather if a significant part of the userbase needs it. So far, I personally don't see a general application. But, It is not up to me to decide this. Its Vigna's project after all.


Lastly, without wanting to be condescending or tell you what to do: IMHO If you are new to unit testing you shouldn't be writing a game engine. Its a tremendously complex project where solutions already exist.

PR #146, opened 2020-03-19
Make find-deps.sh more compatible with other *nixes

astei:
I first ran into this issue when I tried to use the minimization portion of this script as a part of a project I build. The script worked fine on my local system (running Arch Linux) but abjectly failed on the CI system (running FreeBSD). I admit, making portable shell scripts is hard, and I have not yet tackled the task of removing `bash`isms from it (and that's well out of scope for this PR).

This PR fixes the script so that it works on both FreeBSD and Linux (the only two platforms I had conveniently available to test with).

Tested with FreeBSD 12.1-RELEASE and Arch Linux.

The main changes are:

* Fixed `mktemp` invocation so that the syntax is cross-platform compatible.
* Replaced several GNU `sed` usages with `awk` which has better compatibility.

### DIFF
--- a/find-deps.sh
+++ b/find-deps.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #TODO Probably awk or perl can be used in a lot of places
 set -euo pipefail
@@ -117,7 +117,7 @@ case "$command" in
 
   if ! dependencies=$(jdeps -recursive -verbose:class \
     "${classpath_argument[@]+"${classpath_argument[@]}"}" "${analyse_paths[@]}" |\
-    grep '^      -> it\.unimi\.dsi\.fastutil\..*not found$' | sed 's!      -> \(\S*\) \+not found$!\1!' | sort | uniq) \
+    awk '/it\.unimi\.dsi\.fastutil\..*not found$/ { print $2 }' | sort | uniq) \
     || [ -z "$dependencies" ]
   then
     >&2 echo "No unresolved references found - is fastutil on the classpath?"
@@ -158,7 +158,7 @@ case "$command" in
 
   jar_name="$(basename ${dest_path})"
 
-  tmp_dir=$(mktemp -t -d "fastutil-min.XXXX")
+  tmp_dir=$(mktemp -d -t "fastutil-min.XXXX")
   trap "{ rm -rf \"${tmp_dir}\"; exit 255; }" EXIT
 
   ( >&2 echo "Resolving transitive dependencies" )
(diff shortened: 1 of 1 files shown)


incaseoftrouble:
Since I'm sort of responsible for this thing: LGTM. Did know even less about proper bash scripting back then :D

Two ideas: Maybe `declare -a dependencies` and `hash ${jdeps}` -> `command -v "$jdeps" >/dev/null 2>&1`. I recall `command` being more portable than `hash` (same with the other occurence of `hash`). Since you have the tests setup, could you test these changes?

Good to know that someone is using this though :-)

astei:
> Did know even less about proper bash scripting back then :D

I don't really know shell scripting _that_ well, but if `shellcheck` is anything to go by, definitely lots of improvements could be made.

> Two ideas: Maybe `declare -a dependencies` and `hash ${jdeps}` -> `command -v "$jdeps" >/dev/null 2>&1`. I recall `command` being more portable than `hash` (same with the other occurence of `hash`). Since you have the tests setup, could you test these changes?

I'll look into this. Right now I'm trying to get the script working on (of all things) Git Bash for Windows...

develar:
`maven-shade-plugin` is a more robust and easy to use solution. https://github.com/JetBrains/intellij-deps-fastutil/blob/bd5fb6299b00c2474d10c267e2be5c85c8bf0118/pom.xml#L37

incaseoftrouble:
... which requires people to use maven.

develar:
> ... which requires people to use maven.

Well, issue is that in any case you have to use JVM, so, doesn't matter. Especially as in any  bash script doesn't produce consistent results as depends on OS and JVM version (currently, JVM 13 is not supported, for example). In my case on my local machine produced result was 1.7 MB, on jitpack server machine 4 MB. I stopped fighting against bash, and decided to just use maven-based solution.

astei:
That, and as evidenced by this PR, the script is really finicky with anything other than an OS that smells unmistakably like a Linux system.

I think using a new module and using `minimizeJar` works for my particular usecase, but I've been wary of using `minimizeJar` as it has been the source of some pain for me in my previous projects.

incaseoftrouble:
Ah, I misread the maven script. Its sort of independent of the build mechanism of the project, right? This is fine for smaller applications, I guess. The issue is that you need the full dependency tree of the project, not only the src/ directory. For example, if I'm developing `A`, using library `B` in my sourcecode, which in turn has fastutil as transitive runtime dependency, I also need the usages of fast-util in `B` (preferably only those which are reached from the entry points of `A`. In any case, if you are using a proper build system, then shading everything is the simplest way, potentially also minimizing with proguard or similar. I created this script because I was using fastutil in conjunction with a legacy application which uses a `make`-based build system (with handwritten make scripts ...)

I also prefer `minimized-jar`, its a good compromise and in most applications a few MB extra don't really matter.

develar:
> sort of independent of the build mechanism of the project, right?

Yes. In [IDEA](https://github.com/JetBrains/intellij-community) project it is used a regular [dependency](https://github.com/JetBrains/intellij-community/blob/master/.idea/libraries/fastutil_min.xml). 
You specify what do you need in [list](https://github.com/JetBrains/intellij-deps-fastutil/blob/bd5fb6299b00c2474d10c267e2be5c85c8bf0118/src/usages/Usages.java), make a release to trigger build and publishing, and set dependency version in application project.

vigna:
I read in the comments you were still working on the script—progress?

astei:
I gave up trying to make it work on Git Bash for Windows. The changes in this script should still work for making it work better on *BSDs while retaining Linux compatibility.

incaseoftrouble:
food for thought: With WSL now being publicly available with support in major IDEs, I think its fair to resort to that one on windows.

vigna:
Tobias, if you're OK with this I'll merge.

incaseoftrouble:
Yes please, my old script looks extremely ugly :-D

PR #155, opened 2020-05-20
Improved constructors for array-based maps: use fast iterator if possible

develar:
Small correction — for fastutil map fast iterator should be used for performance reasons.

For other map I removed class cast `KEY_OBJ2TYPE` because it is highlighted as unnecessary class cast. 

Resulting java code:

```java
public Object2ObjectArrayMap(final Object2ObjectMap <K,V> m) {
	 this(m.size());
	 int i = 0;
	 ObjectIterator<Object2ObjectMap.Entry <K,V> > iterator = Object2ObjectMaps.fastIterator(m);
	 while(iterator.hasNext()) {
	  Object2ObjectMap.Entry <K,V> e = iterator.next();
	  key[i] = e.getKey();
	  value[i] = e.getValue();
	  i++;
	 }
	}
```

### DIFF
--- a/drv/ArrayMap.drv
+++ b/drv/ArrayMap.drv
@@ -80,7 +80,9 @@ public class ARRAY_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENERIC 
 	public ARRAY_MAP(final MAP KEY_VALUE_GENERIC m) {
 		this(m.size());
 		int i = 0;
-		for(MAP.Entry KEY_VALUE_GENERIC e : m.ENTRYSET()) {
+		ObjectIterator<MAP.Entry KEY_VALUE_GENERIC> iterator = MAPS.fastIterator(m);
+		while(iterator.hasNext()) {
+			MAP.Entry KEY_VALUE_GENERIC e = iterator.next();
 			key[i] = e.ENTRY_GET_KEY();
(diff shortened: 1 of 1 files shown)


vigna:
There's an actual bug, I forgot to set the size. Fixing now.

incaseoftrouble:
> Small correction — for fastutil map fast iterator should be used for performance reasons.

This is not entirely clear btw! The JVM is quite good at figuring out the lifetime of immutable objects, whereas this is quite difficult with mutable objects. Benchmarks would be necessary here. But its not wrong to use fast iterator this way.

vigna:
BTW, that cannot possibly work. If you eliminate the cast, primitive maps won't compile.

vigna:
OK, now you can try the code. It has enhanced constructors for sets and maps. Thanks for the suggestion.

develar:
> This is not entirely clear btw!

Interesting note. From my point of view, if there is API to avoid something, then it should be used. Yes, no doubt, JVM is so heavily optimised, that such simple and logical assumptions are not always correct. But, if I understand correctly, it should be not worse in terms of CPU. In terms of memory and GC...

>  If you eliminate the cast, primitive maps won't compile.

Compiled without any issue (I checked `Short2ShortArrayMap`, and for this class I got IDE warning about unnecessary cast).

PR #158, opened 2020-06-12
New Utility functions for fastutil.

Speiger:
Here some functions that were really needed.
(More the stack then the Priority queue)

Compiled the code to test if it works and added a test to make sure nothing crashes or gives unexpected results.

### DIFF
--- a/drv/AbstractMap.drv
+++ b/drv/AbstractMap.drv
@@ -67,7 +67,22 @@ public abstract class ABSTRACT_MAP KEY_VALUE_GENERIC extends ABSTRACT_FUNCTION K
 	public boolean isEmpty() {
 		return size() == 0;
 	}
-
+		
+#if VALUES_PRIMITIVE
+	/** Adds an increment to value currently associated with a key.
+	 *
+	 * <p>Note that this method respects the {@linkplain #defaultReturnValue() default return value} semantics: when
+	 * called with a key that does not currently appears in the map, the key
+	 * will be associated with the default return value plus
+	 * the given increment.
+	 *
+	 * @param k the key.
+	 * @param incr the increment.
+	 * @return the old value, or the {@linkplain #defaultReturnValue() default return value} if no value was present for the given key.
+	 */
+	public VALUE_GENERIC_TYPE addTo(final KEY_GENERIC_TYPE k, final VALUE_GENERIC_TYPE incr) { throw new UnsupportedOperationException(); }
+#endif
+	
 	/** This class provides a basic but complete type-specific entry class for all those maps implementations
 	 * that do not have entries on their own (e.g., most immutable maps).
 	 *
--- a/drv/AbstractMap.drv
+++ b/drv/AbstractMap.drv
@@ -67,7 +67,7 @@ public abstract class ABSTRACT_MAP KEY_VALUE_GENERIC extends ABSTRACT_FUNCTION K
 	public boolean isEmpty() {
 		return size() == 0;
 	}
-		
+
 #if VALUES_PRIMITIVE
 	/** Adds an increment to value currently associated with a key.
 	 *
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -59,7 +59,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values) {
 		this(values, 0, values.length);
 	}
-	
+
 	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int size) {
 		this(values, 0, size);
 	}
@@ -72,7 +72,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		end = (offset + size) % length;
 		if(length == size) expand();
 	}
-	
+
 	/** Creates a new empty queue with given capacity.
 	 *
 	 * @param capacity the initial capacity of this queue.
@@ -176,7 +176,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		if (start == end) throw new NoSuchElementException();
 		return array[(end == 0 ? length : end) - 1];
 	}
-	
+
 	@Override
 	public KEY_GENERIC_TYPE PEEK(int index) {
 		if(index < 0 || index >= size())
@@ -185,7 +185,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		}
 		return array[(start + index) % length];
 	}
-	
+
 	@Override
 	public void clear() {
 #if KEYS_REFERENCE
@@ -197,22 +197,18 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 #endif
 		start = end = 0;
 	}
-	
-	
+
 	/** Trims the queue to the smallest possible size. */
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	@Override
 	public void trim() {
 		final int size = size();
 #if KEYS_PRIMITIVE
-		final KEY_GENERIC_TYPE[] newArray =	new KEY_GENERIC_TYPE[size + 1];
+		final KEY_GENERIC_TYPE[] newArray = new KEY_GENERIC_TYPE[size + 1];
 #else
-		final KEY_GENERIC_TYPE[] newArray =(KEY_GENERIC_TYPE[])new Object[size + 1];
+		final KEY_GENERIC_TYPE[] newArray = (KEY_GENERIC_TYPE[])new Object[size + 1];
 #endif
-
-		if (start <= end)  {
-			System.arraycopy(array, start, newArray, 0, end - start);
-		}
+		if(start <= end) System.arraycopy(array, start, newArray, 0, end - start);
 		else {
 			System.arraycopy(array, start, newArray, 0, length - start);
 			System.arraycopy(array, 0, newArray, length - start, end);
@@ -221,13 +217,13 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = (end = size) + 1;
 		array = newArray;
 	}
-	
+
 	@Override
 	public int size() {
 		final int apparentLength = end - start;
 		return apparentLength >= 0 ? apparentLength : length + apparentLength;
 	}
-	
+
 	@Override
 	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY()
 	{
@@ -237,17 +233,14 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 #else
 		final KEY_GENERIC_TYPE[] newArray = (KEY_GENERIC_TYPE[])new Object[length];
 #endif
-
-		if (start <= end) {
-			System.arraycopy(array, start, newArray, 0, end - start);
-		}
+		if (start <= end) System.arraycopy(array, start, newArray, 0, end - start);
 		else {
 			System.arraycopy(array, start, newArray, 0, length - start);
 			System.arraycopy(array, 0, newArray, length - start, end);
 		}
 		return newArray;
 	}
-	
+
 	@Override
 	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input)
 	{
@@ -269,7 +262,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		}
 		return input;
 	}
-	
+
 	private void writeObject(java.io.ObjectOutputStream s) throws java.io.IOException {
 		s.defaultWriteObject();
 		int size = size();
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -57,10 +57,20 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	protected transient int end;
 
 	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values) {
+		this(values, 0, values.length);
+	}
+	
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int size) {
+		this(values, 0, size);
+	}
+
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int offset, int size) {
 		if (values.length < 0) throw new IllegalArgumentException("Initial array (" + values.length + ") is negative");
 		array = values;
 		length = values.length;
-		end = length - 1;
+		start = offset;
+		end = (offset + size) % length;
+		if(length == size) expand();
 	}
 	
 	/** Creates a new empty queue with given capacity.
@@ -191,6 +201,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	
 	/** Trims the queue to the smallest possible size. */
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
+	@Override
 	public void trim() {
 		final int size = size();
 #if KEYS_PRIMITIVE
@@ -217,13 +228,14 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		return apparentLength >= 0 ? apparentLength : length + apparentLength;
 	}
 	
+	@Override
 	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY()
 	{
 		final int length = size();
 #if KEYS_PRIMITIVE
-		final KEY_GENERIC_TYPE[] newArray =	new KEY_GENERIC_TYPE[length];
+		final KEY_GENERIC_TYPE[] newArray = new KEY_GENERIC_TYPE[length];
 #else
-		final KEY_GENERIC_TYPE[] newArray =(KEY_GENERIC_TYPE[])new Object[length];
+		final KEY_GENERIC_TYPE[] newArray = (KEY_GENERIC_TYPE[])new Object[length];
 #endif
 
 		if (start <= end) {
@@ -234,6 +246,28 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 			System.arraycopy(array, 0, newArray, length - start, end);
 		}
 		return newArray;
+	}
+	
+	@Override
+	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input)
+	{
+		final int length = size();
+		if(input == null || input.length < length)
+		{
+#if KEYS_PRIMITIVE
+			input = new KEY_GENERIC_TYPE[length];
+#else
+			input = (KEY_GENERIC_TYPE[])new Object[length];
+#endif
+		}
+		if (start <= end) {
+			System.arraycopy(array, start, input, 0, end - start);
+		}
+		else {
+			System.arraycopy(array, start, input, 0, length - start);
+			System.arraycopy(array, 0, input, length - start, end);
+		}
+		return input;
 	}
 	
 	private void writeObject(java.io.ObjectOutputStream s) throws java.io.IOException {
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -65,7 +65,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	}
 
 	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int offset, int size) {
-		if (values.length < 0) throw new IllegalArgumentException("Initial array (" + values.length + ") is negative");
+		if (values.length < size) throw new IllegalArgumentException("Initial array (" + values.length + ") is smaller then the expected size (" + size + ")");
 		array = values;
 		length = values.length;
 		start = offset;
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -154,11 +154,19 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		return array[start];
 	}
 
-
 	@Override
 	public KEY_GENERIC_TYPE LAST() {
 		if (start == end) throw new NoSuchElementException();
 		return array[(end == 0 ? length : end) - 1];
+	}
+	
+	@Override
+	public KEY_GENERIC_TYPE PEEK(int index) {
+		if(index < 0 || index >= size())
+		{
+			throw new NoSuchElementException();
+		}
+		return array[(start + index) % length];
 	}
 
 	@Override
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -56,6 +56,13 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	 *  Might be actually smaller than {@link #start} because {@link #array} is used cyclically. */
 	protected transient int end;
 
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values) {
+		if (values.length < 0) throw new IllegalArgumentException("Initial array (" + values.length + ") is negative");
+		array = values;
+		length = values.length;
+		end = length - 1;
+	}
+	
 	/** Creates a new empty queue with given capacity.
 	 *
 	 * @param capacity the initial capacity of this queue.
@@ -123,11 +130,11 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = newLength;
 	}
 
-	private final void expand() {
+	protected void expand() {
 		resize(length, (int)Math.min(it.unimi.dsi.fastutil.Arrays.MAX_ARRAY_SIZE, 2L * length));
 	}
 
-	private final void reduce() {
+	protected void reduce() {
 		final int size = size();
 		if (length > INITIAL_CAPACITY && size <= length / 4) resize(size, length / 2);
 	}
@@ -168,7 +175,7 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		}
 		return array[(start + index) % length];
 	}
-
+	
 	@Override
 	public void clear() {
 #if KEYS_REFERENCE
@@ -180,18 +187,21 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 #endif
 		start = end = 0;
 	}
-
+	
+	
 	/** Trims the queue to the smallest possible size. */
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	public void trim() {
 		final int size = size();
-		final KEY_GENERIC_TYPE[] newArray =
 #if KEYS_PRIMITIVE
-											new KEY_GENERIC_TYPE[size + 1];
+		final KEY_GENERIC_TYPE[] newArray =	new KEY_GENERIC_TYPE[size + 1];
 #else
-											(KEY_GENERIC_TYPE[])new Object[size + 1];
+		final KEY_GENERIC_TYPE[] newArray =(KEY_GENERIC_TYPE[])new Object[size + 1];
 #endif
-		if (start <= end) System.arraycopy(array, start, newArray, 0, end - start);
+
+		if (start <= end)  {
+			System.arraycopy(array, start, newArray, 0, end - start);
+		}
 		else {
 			System.arraycopy(array, start, newArray, 0, length - start);
 			System.arraycopy(array, 0, newArray, length - start, end);
@@ -200,13 +210,32 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = (end = size) + 1;
 		array = newArray;
 	}
-
+	
 	@Override
 	public int size() {
 		final int apparentLength = end - start;
 		return apparentLength >= 0 ? apparentLength : length + apparentLength;
 	}
+	
+	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY()
+	{
+		final int length = size();
+#if KEYS_PRIMITIVE
+		final KEY_GENERIC_TYPE[] newArray =	new KEY_GENERIC_TYPE[length];
+#else
+		final KEY_GENERIC_TYPE[] newArray =(KEY_GENERIC_TYPE[])new Object[length];
+#endif
 
+		if (start <= end) {
+			System.arraycopy(array, start, newArray, 0, end - start);
+		}
+		else {
+			System.arraycopy(array, start, newArray, 0, length - start);
+			System.arraycopy(array, 0, newArray, length - start, end);
+		}
+		return newArray;
+	}
+	
 	private void writeObject(java.io.ObjectOutputStream s) throws java.io.IOException {
 		s.defaultWriteObject();
 		int size = size();
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -587,6 +587,40 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		return modified;
 	}
 
+#ifdef JDK_PRIMITIVE_PREDICATE	
+	/** Remove from this collection all elements which satisfy the given predicate.
+	 *
+	 * @param filter a predicate which returns {@code true} for elements to be
+	 *        removed.
+	 * @see Collection#removeIf(java.util.function.Predicate)
+	 * @return {@code true} if any elements were removed.
+	 */
+	public boolean removeIf(final JDK_PRIMITIVE_PREDICATE filter) {
+		java.util.Objects.requireNonNull(filter);
+		final KEY_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+#endif
+
+#if KEYS_REFERENCE
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
+		java.util.Objects.requireNonNull(filter);
+		final KEY_GENERIC_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		Arrays.fill(a, j, size, null);
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+#endif
+
 	@Override
 	public KEY_LIST_ITERATOR KEY_GENERIC listIterator(final int index) {
 		ensureIndex(index);
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -588,6 +588,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 * @see Collection#removeIf(java.util.function.Predicate)
 	 * @return {@code true} if any elements were removed.
 	 */
+	@Override
 	public boolean remIf(final JDK_PRIMITIVE_PREDICATE filter) {
 		java.util.Objects.requireNonNull(filter);
 		final KEY_TYPE[] a = this.a;
@@ -598,8 +599,14 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		size = j;
 		return modified;
 	}
-#endif
 
+	@Override
+	@Deprecated
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
+		return remIf((JDK_PRIMITIVE_PREDICATE) key -> filter.test(KEY2OBJ(KEY_NARROWING(key))));
+	}
+#else
+	@Override
 	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
 		java.util.Objects.requireNonNull(filter);
 		final KEY_GENERIC_TYPE[] a = this.a;
@@ -613,6 +620,8 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		size = j;
 		return modified;
 	}
+
+#endif
 
 	@Override
 	public KEY_LIST_ITERATOR KEY_GENERIC listIterator(final int index) {
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -588,7 +588,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 * @see Collection#removeIf(java.util.function.Predicate)
 	 * @return {@code true} if any elements were removed.
 	 */
-	public boolean removeIf(final JDK_PRIMITIVE_PREDICATE filter) {
+	public boolean remIf(final JDK_PRIMITIVE_PREDICATE filter) {
 		java.util.Objects.requireNonNull(filter);
 		final KEY_TYPE[] a = this.a;
 		int j = 0;
@@ -600,19 +600,19 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	}
 #endif
 
-#if KEYS_REFERENCE
 	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
 		java.util.Objects.requireNonNull(filter);
 		final KEY_GENERIC_TYPE[] a = this.a;
 		int j = 0;
 		for(int i = 0; i < size; i++)
 			if (!filter.test(a[i])) a[j++] = a[i];
+#if KEYS_REFERENCE
 		Arrays.fill(a, j, size, null);
+#endif
 		final boolean modified = size != j;
 		size = j;
 		return modified;
 	}
-#endif
 
 	@Override
 	public KEY_LIST_ITERATOR KEY_GENERIC listIterator(final int index) {
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -424,15 +424,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	public boolean isEmpty() {
 		return size == 0;
 	}
-
-	/** Trims this array list so that the capacity is equal to the size.
-	 *
-	 * @see java.util.ArrayList#trimToSize()
-	 */
-	public void trim() {
-		trim(0);
-	}
-
+	
 	/** Trims the backing array if it is too large.
 	 *
 	 * If the current array length is smaller than or equal to
@@ -449,6 +441,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 */
 
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
+	@Override
 	public void trim(final int n) {
 		// TODO: use Arrays.trim() and preserve type only if necessary
 		if (n >= a.length || size == a.length) return;
--- a/drv/Collection.drv
+++ b/drv/Collection.drv
@@ -157,7 +157,7 @@ public interface COLLECTION KEY_GENERIC extends Collection<KEY_GENERIC_CLASS>, K
 	@Deprecated
 	@Override
 	default boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
-		return removeIf((JDK_PRIMITIVE_PREDICATE) key -> filter.test(KEY2OBJ(KEY_NARROWING(key))));
+		return remIf((JDK_PRIMITIVE_PREDICATE) key -> filter.test(KEY2OBJ(KEY_NARROWING(key))));
 	}
 
 	/** Remove from this collection all elements which satisfy the given predicate.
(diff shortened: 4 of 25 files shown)


Speiger:
If I find more "Random" functions that are missing I will add them as they were being found, though if this gets accepted I will just make a new pull request.
(Stuff like clear)

Speiger:
List of contents for simplification:
- Tutorial on how to generate the source-code on windows
- Stack.class gets a clear function because you sometimes want to clear a stack instead of pouring it out.
- PriorityQueue gets a peek function.
- PriorityQueue gets a toArray() function / array constructor for serialization, because if you don't use the default java serialization this class becomes useless otherwise.
- PriorityQueue expand/shrink function is protected now since it is very often a target for custom implementation to prevent growth/shrinkage.
- Object2Number.class gets the addTo function and all its implementations get handled too.
- Added tests for peek and clear function, and added missing addTo function to testInt2IntMap.

@vigna anything that is right out of the gate a no go for you?

Speiger:
Also if needed i can make subbranches for everything.

drv/AbstractMap.drv
```diff
@@ -67,7 +67,22 @@ public abstract class ABSTRACT_MAP KEY_VALUE_GENERIC extends ABSTRACT_FUNCTION K
...
+	 * <p>Note that this method respects the {@linkplain #defaultReturnValue() default return value} semantics: when
+	 * called with a key that does not currently appears in the map, the key
+	 * will be associated with the default return value plus
+	 * the given increment.
+	 *
+	 * @param k the key.
+	 * @param incr the increment.
+	 * @return the old value, or the {@linkplain #defaultReturnValue() default return value} if no value was present for the given key.
+	 */
+	public VALUE_GENERIC_TYPE addTo(final KEY_GENERIC_TYPE k, final VALUE_GENERIC_TYPE incr) { throw new UnsupportedOperationException(); }
```
incaseoftrouble, line 83: Isn't this a special case of merge / compute?

drv/ArrayFIFOQueue.drv
```diff
@@ -192,13 +210,32 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = (end = size) + 1;
 		array = newArray;
```
incaseoftrouble, line 211: instead: array = TO_KEY_ARRAY(), right? Looks like its exactly the same code (+ changes to start, length and end)

drv/ArrayList.drv
```diff
@@ -587,6 +587,40 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
+#endif
+
+#if KEYS_REFERENCE
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
+		java.util.Objects.requireNonNull(filter);
+		final KEY_GENERIC_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		Arrays.fill(a, j, size, null);
```
incaseoftrouble, line 617: This can be omitted I guess. Everything in a after `size` will not be read.

src/it/unimi/dsi/fastutil/Stack.java
```diff
@@ -52,7 +52,12 @@
 	 */
 
 	boolean isEmpty();
-
+	
+	/**
+	 * Clears the Contents of the stack
```
incaseoftrouble, line 57: *`contents` :-)

incaseoftrouble:
I added some small comments but looks good to me overall. Regarding the tutorial: IMO WSL should be the new default (after all, it has proper support by Microsoft), however if you got it to work with cygwin, thats fine too.

drv/AbstractMap.drv
Speiger: Kinda, but this is the object allocation free variant and only works with numeric values since it is counting, and if it is not present then it just adds it with 0 delay.

Compute/Merge is basically slower due to the lambda.

drv/ArrayFIFOQueue.drv
Speiger: First toArray has to be accepted by @vigna but that kinda makes sense. Though custom implementations would be more dangerous then.

drv/ArrayList.drv
Speiger: No, the clear function even clears the array too. Otherwise objects won't be de-allocated by java if there is a reference left.
for primitives yes you can leave them, but this is the Object version only.

src/it/unimi/dsi/fastutil/Stack.java
Speiger: ty will be patched.

Speiger:
@incaseoftrouble thanks for the review, i just got the tutorial with cygwin explained by someone else and that's why I focused on that.

drv/AbstractMap.drv
incaseoftrouble: Overhead of e.g., `Double::sum` vs `a + b` very likely is negligible I'd guess. However, I think there aren't primitive binary operators so there would be boxing...

Speiger: Eh fastUtil add comparators themselves.

Speiger: Just to give some perspective @incaseoftrouble:
image
That is the compare code and the results from a JUnit4 Benchmark are these:
```
MiscTest.compareList: [measured 1000 out of 1100 rounds, threads: 1 (sequential)]
 round: 0.01 [+- 0.00], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 14, GC.time: 0.01, time.total: 11.26, time.warmup: 1.13, time.bench: 10.14

MiscTest.compareListLambda: [measured 1000 out of 1100 rounds, threads: 1 (sequential)]
 round: 0.01 [+- 0.00], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 19, GC.time: 0.01, time.total: 12.50, time.warmup: 1.23, time.bench: 11.28
```
It is only a tiny differences but some factors you can see is: GC is running more and the code runs a tiny bit slower.
instead of 10,14ms per execution it takes a full 1,14ms longer to execute the same code with the same seed just because a lambda was added. Rule of thumb is, the less a lambda is used or created the better it is. That is why I rather promote non lambda solutions.

incaseoftrouble: Yes, it is a 10% difference (even though I think the setup of the test should be separated), which is even less then I expected (its barely significant). Of course lambdas come with a cost. But thats not why lambdas were introduced in the first place.

From a practical perspective: Its a 1ms difference on a list with ~100k elements (so something like 1 million calls, assuming `n log n`). So, this hardly makes a difference (if you are dealing with anything where a difference of 1ms for 1 million calls makes a difference you should probably be writing in assembler). Maintaining an additional function with special semantics is a real difference. You could probably spend time in *any* other part of this library and optimize a bit to obtain a more noticeable effect. In my eyes, this is the prototical "premuature optimization" Knuth warned about.

incaseoftrouble: Btw AFAIK method refence lambdas are handled differently than anonymous classes. So `List.sort(Integer::compare)` should be noticeably faster than `List.sort((a, b) -> b - a)`

Speiger: Note right out of the gate: I know why lambdas exist, and I see their use cases and use them myself.
But they should never be the priority.
About the performance gain. You are aware that Integer::compare is just ((k, v) -> Integer.compare(k, v)) it is still a allocation.
It can be compared to a += b or a = a + b;
And since lambdas are rarely cache-able (speaking from experience) and the 10% lag comes from how often they are used in a single allocation what do you think does this to performance if you don't have a 1-10ms task, because you are reallocation the same function every time it is called.
So overhead rises with less executions per call.

Simple sum up from me: Lambdas are useful if performance isn't a factor, UI interaction, rarely called lazyloads, or some other cases. But otherwise avoid using them, because you could run a lot more code otherwise.

Also I hope I do not sound rude, i find this discussion really interesting.

Speiger: Also why i added this "addTo" function to the interface was so you could use object2IntMap instead of Object2ObjectOpenHashMap to access that function. Just to remove unnecessary casts or forced breaking of the rule that you should use interfaces on the left.

incaseoftrouble: >  About the performance gain. You are aware that Integer::compare is just ((k, v) -> Integer.compare(k, v)) it is still a allocation.

Technically, these two are quite different under the hood. The JVM can optimize direct method references much easier (see, e.g. [here](https://softwareengineering.stackexchange.com/questions/277473/is-there-a-performance-benefit-to-using-the-method-reference-syntax-instead-of-l)).

> It can be compared to a += b or a = a + b;

These two are the same (see e.g. [here](https://stackoverflow.com/questions/39821936/performance-of-in-java)). Compilers are incredibly smart.

>  So overhead rises with less executions per call.

Yes and no. The *absolute* overhead of a lambda is minimal.

> Simple sum up from me: Lambdas are useful if performance isn't a factor, UI interaction, rarely called lazyloads, or some other cases. But otherwise avoid using them, because you could run a lot more code otherwise.

They are also useful for core performance areas, called hundreds of times per second. Simply because you save so much effort and make code maintainable so that you can spend effort on where it actually matters. Just see the example, even though there are 1 million (!) calls to the lambda, the execution takes a few milliseconds. I'd bet money that in basically every single application where the `addTo` method is used, a (primitive) lambda together with some optimization elsewhere has a larger impact.

Given that you have infinite time, its of course most performant to not use lambdas at all and hand tailor every single datastructure. However, usually you don't have infinite time, so prioritizing clean, readable code actually yields more performant code in practice. Typically, you'd probably write your code reasonably efficient, the run a profiler and identify hotspots and only then perform these fine-grained optimizations at these particular hotspots (at least that's what I have learned over the years). This is even more important for libraries. The larger the public API the larger the maintenance burden.

> forced breaking of the rule that you should use interfaces on the left

This is fine to do in particular scenarios. For example, Guava advocates to use the Immutable collection variants wherever they are used. This rule essentially is an instance of "least required knowledge" principle. Usually, you don't need to know how exactly the map is implemented, you only need some properties of it. This is what interfaces are for. However, in some cases you might need implementation details and its perfectly fine to specialize down to the implementation.


To summarize: Yes, lambdas have some overhead. However, it doesn't matter in practically any application. Even for, e.g., programming contests, where performance is the only concern, nobody cares whether you use lambdas or not (well, to be fair, most people use C++ there ...) So just introducing new API just to avoid using lambdas is not good (IMHO!). That is what they are made for, after all. However, if there are other legitimate reasons I won't oppose this :-)

Speiger: > Technically, these two are quite different under the hood. The JVM can optimize direct method references much easier (see, e.g. here).

Interesting. Good to know. Thanks.

> Yes and no. The absolute overhead of a lambda is minimal.

If you do one or two, but lets say you are stacking these everywhere instead of normal parameters. Stuff like print(String) to printString(Supplier<String>) that calls get instantly. Or use layers of lambda after lambda (have seen 5 layers in worst case scenarios) it can have a performance impact.

One game example that I love to show like that is: Minecraft, they wen't nuts on Lambdas and started using them excessively everywhere even if it is not even reasonable to use them for their quality. The game runs since then on the server side around 50% slower then previous iterations of the game. Now lambdas are not the fault maker at here, but they are a big factor next to the design of the code in general.

> They are also useful for core performance areas, called hundreds of times per second. Simply because you save so much effort and make code maintainable so that you can spend effort on where it actually matters. Just see the example, even though there are 1 million (!) calls to the lambda, the execution takes a few milliseconds. I'd bet money that in basically every single application where the addTo method is used, a (primitive) lambda together with some optimization elsewhere has a larger impact.
Given that you have infinite time, its of course most performant to not use lambdas at all and hand tailor every single datastructure. However, usually you don't have infinite time, so prioritizing clean, readable code actually yields more performant code in practice. Same goes for the whole of the stream() API. Typically, you'd probably write your code reasonably efficient, the run a profiler and identify hotspots and only then perform these fine-grained optimizations at these particular hotspots (at least that's what I have learned over the years). This is even more important for libraries. The larger the public API the larger the maintenance burden.

Yeah I agree with you on that point, readable code leads to better performance, and lambdas can be used in high performance way even at a smaller loss but not if they are created 100k times for 1M objects to compare (ignore primitives right now and i mean 1 comparator is created per 10 objects to sort). Giving my Minecraft example.

But you can create readable code even without lambdas. It depends on your design. Lambda basically reduces the gap of skill required making readable code. Which is not a bad thing, but also not a good thing, its a thing.

> This is fine to do in particular scenarios. For example, Guava advocates to use the Immutable collection variants wherever they are used. This rule essentially is an instance of "least required knowledge" principle. Usually, you don't need to know how exactly the map is implemented, you only need some properties of it. This is what interfaces are for. However, in some cases you might need implementation details and its perfectly fine to specialize down to the implementation.

Fully agree there, just in this case that was like Collection would contain the contains method and you had to go to the next level to execute it.

> To summarize: Yes, lambdas have some overhead. However, it doesn't matter in practically any application. Even for, e.g., programming contests, where performance is the only concern, nobody cares whether you use lambdas or not (well, to be fair, most people use C++ there ...) So just introducing new API just to avoid using lambdas is not good (IMHO!). That is what they are made for, after all. However, if there are other legitimate reasons I won't oppose this :-)

"Most using C++" yeah, but the compilation process is different there but whatever.
It really depends on how often it is called. You can compare it honestly to Math.sqrt vs not to use it for distance checks.
And about not introducing a new api, it really depends. In this case this gets rid of the lazyload need in general, it results in cleaner code and faster code too since either special interfaces not being required or boxing/unboxing not being an issue.

If a better result could be achieved with a lambda for example (or lets say functional interface) then yeah go ahead, because it makes sense to get something better.

It really depends on how to add it. Which reminds me i need to make another optimization to fastutil that i noticed could save a lot of time while being almost a no change at all. You see in a minute hopefully

incaseoftrouble: Agree with most things. My main point is: Lambdas aren't bad per se, and (somehwat) excessively using them is fine for development. Once you are sure that lambdas are the problem, start to replace them.

Regarding the minecraft example: I have also worked with that code base. Many plugins were just awfully designed in terms of algorithmics (which is fine typically, because minecraft's server side doesn't have that much throughput on smaller servers). Stuff like using `LinkedList` where random access is needed, `TreeMap` where no sorting is required, using String concatenation instead of `StringBuilder`, etc. Or simply the wrong algorithmic approach to begin with. This is much more impactful than lambdas. And such issues are present in 99% of all codebases, I am very sure, especially in something accessible like minecraft.

What would be an elegant alternative here: Providing `merge` / `compute` primitive variants where it makes sense. For example `map.merge(key, value, Double::sum)`, where the last is a `DoubleBinaryOperator` (this exists in standard Java API for the "base" primitives int, long, and double). This would work for all `2Int`, `2Long`, and `2Double` maps (most other primitives don't really make sense in most cases anyway, since the JVM only really supports instructions for these three). It would also work for the "smaller" types by using "safe" casts (I think I once added similar functionality for `compute` if I recall correctly...). I'd really be interested in the performance difference of this approach vs. a direct `addTo` method. Its probably <= 20% and I am really inclined to take that penalty for the sake of flexibility (what if you'd like to instead `divideBy` or `multiplyWith`?).

> "Most using C++" yeah, but the compilation process is different there but whatever.

No? Depending on where you go, settings may be different, but, for example, ICPC uses `g++ -g -O2 -std=gnu++17 -static ${files}` and `javac -encoding UTF-8 -sourcepath . -d . ${files}`.

Speiger: > Agree with most things. My main point is: Lambdas aren't bad per se, and (somehwat) excessively using them is fine for development. Once you are sure that lambdas are the problem, start to replace them. I think we agree on that.

I can stand next to that.

> Regarding the minecraft example: I have also worked with that code base. Many plugins were just awfully designed in terms of algorithmics (which is fine typically, because minecraft's server side doesn't have that much throughput on smaller servers). Stuff like using LinkedList where random access is needed, TreeMap where no sorting is required, using String concatenation instead of StringBuilder, etc. Or simply the wrong algorithmic approach to begin with. This is much more impactful than lambdas. And such issues are present in 99% of all codebases, I am very sure, especially in something accessible like minecraft.

Interesting, yeah i can see that but if you have seen chunkloading in 1.14 or newer you will fall appart. It feels like every 3rd line of code is a new lambda on a 600 line code file. It got excessive. That's why I think its one of the reasons it runs slow. About string though. Stuff like "text"+value+"Other" text is not a bad thing in java for the whole reason: Any complex string that is created that way gets in the compilation process to byte code converted to a stringbuilder automatically. That is one of the only code changes that the compiler does that i am aware of. So seeing it in src that way means nothing. And the compiler is a good judge so far from what I have seen.

> What would be an elegant alternative here: Providing merge / compute primitive variants where it makes sense. For example map.merge(key, value, Double::sum)...........

Yeah i would be interested to see that too. Maybe i am going to write a speed test for that later.
But addTo is for one specific replacement that I have used so often that it makes sense over basic algorithmic calculation.
Before i would do this in my most optimized version:
```java
Map<Object, Integer> map = null;//Normal Hashmap

public void addStuff(Object obj, int adding)
{
	if(map.containsKey(obj))
	{
		map.put(obj, map.get(obj)+adding);
		return;
	}
	map.put(obj, adding);
}
```
That was when i was a beginner. Later it got converted to something like this:
```java
Map<Object, MutableInt> map = null;//Normal Hashmap

public void addStuff(Object obj, int adding)
{
	MutableInt value = map.get(obj);
	if(value == null)
	{
		value = new MutableInt();
		map.put(obj, value);
	}
	value.add(adding);
}
```
All that got replaced by the addToFunction because it was no longer necessary. 
It just simplifies the code so much and provides a simple mapped counter, which has so many uses.
One of the mods that I write for example is a 3D energynetwork grid that keeps track of how much energy was send over a given time. That uses the addToFunction, or shocks Entities when they collide with uninsulated cables, since it could be multiple cables it collides with at the same time it needs to combine those. Object2IntMap<Entity>,
Its rarely the case that you need multiply or division. And even if then the compute function exists to fill the gap.
But you need a mapped counter more often then anything.

> No? Depending on where you go, settings may be different, but, for example, ICPC uses g++ -g -O2 -std=gnu++17 -static ${files} and javac -encoding UTF-8 -sourcepath . -d . ${files}.

you missunderstood me here xD (my fault). I meant that lambdas in C++ are a different thing then in java. The compilation process results in 2 different results of the same named thing. Like language false friends. Like the word Die in english vs german.
In English Die means death (simplified)
In German Die is the female version of the word "The"

incaseoftrouble: > Any complex string that is created that way gets in the compilation process to byte code converted to a stringbuilder automatically.

Only if its a static string. Not if you are building a string in a loop or anything dynamic (then its a lot of copying) since the compiler then can't prove equivalence in general.

> example / replaced by addTo

Yes, and `map.merge(key, val, Integer::sum)` also does exactly that :-) `Map#merge` was introduced exactly for this type of applications! But even then, using a specialized addTo can only give you a benefit of at most x2 (instead of two hash lookups you only need one, but that's the only advantage).

This probably is faster than using `Object2IntMap` without merge or similar tricks due to Java's special handling of small Integer objects.

> Not using division often

Well, no, it just depends on the application area. If you are working with, e.g., discrete distributions, `Object2DoubleMap` is a natural representation and division / multiplication is a natural operation for, e.g., reweighting.

I just ran some JMH tests. `addTo` is only slightly faster than a `merge` with an `IntUnaryOperator` (~5% on a quick benchmark). Its also only barely faster than a "naive" approach (see below). However, all three variants manage ~80000 insertions per millisecond (!) on a single core. Note: I used `Long` as key, these have a very quick hashcode, but whatever objects you are dealing with probably has a cached hash code anyway.

Numbers:
```
Benchmark                     Mode  Cnt   Score   Error  Units
Test.runFastutilAddTo        thrpt    2  82.331          ops/s
Test.runFastutilCustomMerge  thrpt    2  79.542          ops/s
Test.runFastutilMerge        thrpt    2  63.108          ops/s
Test.runFastutilPlain        thrpt    2  77.797          ops/s
Test.runMerge                thrpt    2  53.676          ops/s
Test.runPlain                thrpt    2  53.821          ops/s
```
CustomMerge is with `IntUnaryOperator`, Merge with normal `BiFunction`, Plain with
```java
      Integer previous = map.get(key);
      if (previous == null) {
        map.put(key, value);
      } else {
        map.put(key, previous + value);
      }
```
(and similar for fastutil). 1000000 insertions per round. The difference between plain and merge becomes more apparent for more expensive `hashCode` of the key objects.

Speiger: > Only if its a static string. Not if you are building a string in a loop or anything dynamic (then its a lot of copying) since the compiler then can't prove equivalence in general. Just run something like.

Eh just made a jar with your code and this is what the bytecode pumps out. WHILE the human readable code looks like this:
ByteCode
```java
     public test() { //()V
         <localVar:index=2 , name=i , desc=I, sig=null, start=L1, end=L2>
         <localVar:index=0 , name=this , desc=Lspeiger/src/game/GameManager;, sig=null, start=L3, end=L4>
         <localVar:index=1 , name=a , desc=Ljava/lang/String;, sig=null, start=L5, end=L4>

         L3 {
             ldc "" (java.lang.String)
             astore1
         }
         L5 {
             iconst_0
             istore2
         }
         L1 {
             iload2
             ldc 1000000 (java.lang.Integer)
             if_icmpge L2
         }
         L6 {
             new java/lang/StringBuilder
             dup
             invokespecial java/lang/StringBuilder <init>(()V);
             aload1
             invokevirtual java/lang/StringBuilder append((Ljava/lang/String;)Ljava/lang/StringBuilder;);
             ldc "." (java.lang.String)
             invokevirtual java/lang/StringBuilder append((Ljava/lang/String;)Ljava/lang/StringBuilder;);
             invokevirtual java/lang/StringBuilder toString(()Ljava/lang/String;);
             astore1
         }
         L7 {
             iinc 2 1
             goto L1
         }
         L2 {
             return
         }
         L4 {
         }
     }
```
Java Code (Decompiled with fernflower):
```java
    public void test() {
        String a = "";
        for (int i = 0; i < 1000000; ++i) {
            a += ".";
        }
    }
```

The code is super un-optimized but it is converted to a stringbuilder. But my statement is still true. It is converted to a string.

But lets  give a example with a test case like this:
Java Code (Source)
```java
	public void subTest(Object key, String test, int result)
	{
		String s = "Totally not testing string builder["+key+"] functionality"+result+test;
	}
```
Java Code (Decompiled with fernflower):
```java
   public void subTest(Object key, String test, int result) {
      (new StringBuilder()).append("Totally not testing string builder[").append(key).append("] functionality").append(result).append(test).toString();
   }
```
Byte Code:
```java
     public subTest(java.lang.Object arg0, java.lang.String arg1, int arg2) { //(Ljava/lang/Object;Ljava/lang/String;I)V
         <localVar:index=0 , name=this , desc=Lspeiger/src/game/GameManager;, sig=null, start=L1, end=L2>
         <localVar:index=1 , name=key , desc=Ljava/lang/Object;, sig=null, start=L1, end=L2>
         <localVar:index=2 , name=test , desc=Ljava/lang/String;, sig=null, start=L1, end=L2>
         <localVar:index=3 , name=result , desc=I, sig=null, start=L1, end=L2>
         <localVar:index=4 , name=s , desc=Ljava/lang/String;, sig=null, start=L3, end=L2>

         L1 {
             new java/lang/StringBuilder
             dup
             invokespecial java/lang/StringBuilder <init>(()V);
             ldc "Totally not testing string builder[" (java.lang.String)
             invokevirtual java/lang/StringBuilder append((Ljava/lang/String;)Ljava/lang/StringBuilder;);
             aload1 // reference to arg0
             invokevirtual java/lang/StringBuilder append((Ljava/lang/Object;)Ljava/lang/StringBuilder;);
             ldc "] functionality" (java.lang.String)
             invokevirtual java/lang/StringBuilder append((Ljava/lang/String;)Ljava/lang/StringBuilder;);
             iload3
             invokevirtual java/lang/StringBuilder append((I)Ljava/lang/StringBuilder;);
             aload2 // reference to arg1
             invokevirtual java/lang/StringBuilder append((Ljava/lang/String;)Ljava/lang/StringBuilder;);
             invokevirtual java/lang/StringBuilder toString(()Ljava/lang/String;);
             astore4
         }
         L3 {
             return
         }
         L2 {
         }
     }
```

As you can see even the decompiler without bytecode is showing me a StringBuilder and the byte code shows it too.
If a String is shown as a String can not really be trusted in decompiler because it might as well be a stringbuilder in the actual bytecode and thats the one thats being interpreted, not the human readable source.

> I just ran some JMH tests.

Yeah roughly what i expected, but really interesting data though. I didn't expect the boxing/unboxing method to be so fast though.

incaseoftrouble: > The code is super un-optimized but it is converted to a stringbuilder. But my statement is still true. It is converted to a StringBuilder, but it results in a string.

No. The Stringbuilder is used for a single `+=` and then converted back to an (immutable) String, L6 is the loop body. Just run the code. Its orders of magnitude slower. Appending like this is `O(n^2)` while direct stringbuilder is `O(n)`.

> Yeah roughly what i expected, but really interesting data though. I didn't expect the boxing/unboxing method to be so fast though.

Integers are a special case due to the small integer cache, but boxing is indeed pretty fast (its a natural hotspot method).

Speiger: > No. The Stringbuilder is used for a single += and then converted back to an (immutable) String, L6 is the loop body. Just run the code. Its orders of magnitude slower. Appending like this is O(n^2) while direct stringbuilder is O(n).

My Point was that building a string is not always done via String.contact, but with string builders. Ofc in your scenario it uses the most optimal way to do it, but any string that is generated like how i showed it in my example is at least optimized.

> Integers are a special case due to the small integer cache, but boxing is indeed pretty fast (its a natural hotspot method).

I know about  the cache etc, but i remember when upgrading to fastutil it was much faster speed compared to the boxing/unboxing. though I think GC had a role in there too.

Speiger:
@vigna this pull request is finished from my side. Now basically approval process is waiting.

Feature list:
- Tutorial on how to generate the source-code on windows
- Added tests for peek and clear function, and added missing addTo function to testInt2IntMap.
- Stack.class gets a clear function because you sometimes want to clear a stack instead of pouring it out.
- PriorityQueue:
   - peek function
   - trim function now accessible through interface (all implementations used it)
   - toArray functions so they can be saved/serialized/converted as needed and can be used outside of the java serialization scope.
   - FIFOArray gets constructors that allow to provide a array. Trusting the user in that regard.
   - FIFOArray grow/shrink function are protected now since they are often the target of customization.
- Maps: 
   - Object2Number interface gets now access to the addToFunction.
   - Object2Object getOrDefault & putIfAbsent get their Optimizations enabled.
- Lists:
   - Reverse function added to Lists class.
   - trim function is now accessible through the interface.
   - removeIf got optimized to be faster.
   - size function was still allowed in the un-modifiable list. (It changes the array contents)

@incaseoftrouble do you want to look over the changes again to see if i made mistakes somewhere?

vigna:
Well, there was a bug you just fixed. Do unit test cover 100% the new code?

Speiger:
@vigna deciding factor for me was: Do you use the trim function more frequently then frontcoded lists.
And same could apply to the PriorityQueue last function or any other function that is specifically not for frontcodedLists.

Anyways added tests for functions that needed them. (Anything that is not a interface link to a existing function with test)

One somewhat major change i did to List/Collection was make primitive functions from removeIf to remIf that makes it more consistent with the rem function, but also fixes the required casting for lambdas if the function was used.

Speiger:
@vigna anything left that would stop this pull from being accepted?

vigna:
I have to check everything, carefully. Just that.

Speiger:
Ok just wanted to make sure that I didn't forget to do something (before you did your careful check)

drv/ArrayFIFOQueue.drv
```diff
@@ -56,6 +56,23 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
...
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values) {
+		this(values, 0, values.length);
+	}
+	
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int size) {
+		this(values, 0, size);
+	}
+
+	public ARRAY_FIFO_QUEUE(final KEY_GENERIC_TYPE[] values, int offset, int size) {
+		if (values.length < 0) throw new IllegalArgumentException("Initial array (" + values.length + ") is negative");
```
incaseoftrouble, line 68: arrays can't have negative size. Rather check if size < 0 or size > values.length I guess

drv/ArrayFIFOQueue.drv
```diff
@@ -123,11 +140,11 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = newLength;
 	}
 
-	private final void expand() {
+	protected void expand() {
```
incaseoftrouble, line 143: Not sure if this should be overrideable. Its a rather internal functionality.

drv/ArrayList.drv
```diff
@@ -587,6 +580,40 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
 
+#ifdef JDK_PRIMITIVE_PREDICATE	
+	/** Remove from this collection all elements which satisfy the given predicate.
+	 *
+	 * @param filter a predicate which returns {@code true} for elements to be
+	 *        removed.
+	 * @see Collection#removeIf(java.util.function.Predicate)
+	 * @return {@code true} if any elements were removed.
+	 */
+	public boolean remIf(final JDK_PRIMITIVE_PREDICATE filter) {
```
incaseoftrouble, line 591: Maybe `removeIfPrimitive`?

drv/ArrayList.drv
```diff
@@ -587,6 +580,40 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+#endif
+
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
```
incaseoftrouble, line 603: add deprecation note

drv/ArrayList.drv
```diff
@@ -587,6 +580,40 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
+		java.util.Objects.requireNonNull(filter);
+		final KEY_GENERIC_TYPE[] a = this.a;
```
incaseoftrouble, line 605: should delegate to the primitive variant above if `filter` is instance of a primitive filter.

drv/ArrayList.drv
incaseoftrouble: Actually, nevermind. This should be `removeIf`, the primitive predicates do not extend the generic predicates, so there is no clash

drv/Collection.drv
```diff
@@ -167,8 +167,7 @@ public interface COLLECTION KEY_GENERIC extends Collection<KEY_GENERIC_CLASS>, K
 	 * @see Collection#removeIf(java.util.function.Predicate)
 	 * @return {@code true} if any elements were removed.
 	 */
-	@SuppressWarnings("overloads")
-	default boolean removeIf(final JDK_PRIMITIVE_PREDICATE filter) {
+	default boolean remIf(final JDK_PRIMITIVE_PREDICATE filter) {
```
incaseoftrouble, line 170: should be `removeIf` (this change would break public API!)

drv/List.drv
```diff
@@ -99,7 +99,31 @@ public interface LIST KEY_GENERIC extends List<KEY_GENERIC_CLASS>, COLLECTION KE
...
+	
+	/** Trims this array list so that the capacity is equal to the size.
+	 *
+	 * @see java.util.ArrayList#trimToSize()
+	 */
+	default void trim() {
+		trim(0);
+	}
 
+	/** Trims the backing array if it is too large.
```
incaseoftrouble, line 111: Backing array makes no sense here. What about e.g. linked lists?

drv/Lists.drv
```diff
@@ -46,6 +46,16 @@ public final class LISTS {
 		}
 		return l;
 	}
+	
+	public static KEY_GENERIC LIST KEY_GENERIC reverse(final LIST KEY_GENERIC l) {
+		int size = l.size();
+		for (int i=0, mid=size>>1, j=size-1; i<mid; i++, j--) {
```
incaseoftrouble, line 52: This should be done with a view I think (see, e.g. Guavas Lists#reverse) for `O(1)` reversal instead of `O(n)`.

incaseoftrouble (commented):
My opinions:
 * `trim()` does not belong in the public interface. Maybe something like `Lists.trimIfPossible()` if necessary. In general I would adhere to Java's collection API style
 * Still don't like `addTo` :-)
 * There are a lot of disjoint changes in here. It might make sense to split up the MR into smaller bits
 * Many whitespace changes and differences in codestyle which should be fixed.

drv/ArrayFIFOQueue.drv
Speiger: Yeah, I can see and agree why that is a better idea.

drv/ArrayFIFOQueue.drv
Speiger: True, but I have seen a lot of cases where this function or the shrink function were the target of overrides.
So instead of overriding 3 functions that all call this exact function (which they want to modify in the first place) it is overriding the function directly. resize is still private.

drv/ArrayList.drv
Speiger: IntList.rem() to remove a primitive. So it is consistent with lists.
And when i tried using removeIf I have to cast because otherwise the compiler will loses its mind.
In that case it is better to split the function names from each other.

drv/ArrayList.drv
Speiger: Ty

drv/ArrayList.drv
Speiger: I am not sure if the compiler likes that, but on the todolist it is.

drv/Collection.drv
Speiger: But that fixes the "@SuppressWarnings("overloads")" bug that was present in the first place where casting was required.

drv/List.drv
Speiger: Wait they exist in fastUtil?

drv/Lists.drv
Speiger: Guavas implementation is a Lazy implementation, with a ObjectAllocation where you lose the original List Reference.
That just adds a layer on top of the existing one.
This is the direct implementation and should be O(1) or O/2 since it stops at the half point.

Also yes a lazy implementation can make sense if you don't want to modify the original, but the intend of this method is to modify the original. Could be compared to List.sort using Lists.sort where it gives you a sorted list that does not modify the original.

Speiger:
@incaseoftrouble 
- 1: My argument here is Collection.sort(List<> list) implementation is: list.sort(null/comparator), and not sure if people would look for that, or they would directly cast which is a huge issue in fastutil where casting for every "new" and "useful" feature you have to cast or break javas coding recommendation of using interfaces instead of direct classes where present. (ObjectArrayList<> list = new ObjectArrayList<>();
- 2: I know :-)
- 3: I can kinda agree with the disjointment, but the core problem is all the same here, functions that are essential but missing. (Pull your left arm around your right ear to scratch the left ear). Though in the future my goal is to have more dedicated patches.
- 4: I tried to mimic it as best as possible, but there is no real rulesheet here to go of on or add to the workspace. Though adding that is not my job. Could you point out exact issues so I can address them?
- 5: Patch is coming.

drv/ArrayList.drv
incaseoftrouble: Up for @vigna to decide.

drv/ArrayFIFOQueue.drv
incaseoftrouble: Something @vigna should decide IMO

drv/Collection.drv
incaseoftrouble: How is that a bug? You can just write `(int arg) -> arg < 0` instead of `(IntPredicate) arg -> arg < 0` (or something like that) to disambiguate the call btw. Breaking public API is not good.

drv/List.drv
incaseoftrouble: They don't I think, but someone might implement them subclassing this interface. That's the point, its a generic list collection interface, not a array list interface.

drv/Lists.drv
incaseoftrouble: This implementation is `O(n)` :-) 

The problem is that this method assumes the given list is random access (which may not always be the case). At least some sort of warning-documentation should be added then.

incaseoftrouble:
1) That's what documentation is for :-)

3) Well that depends on your use case, I haven't missed them so far ;-) But splitting up such features is mainly intended to be easier to review (I know from experience that splitting up your changes in smaller bits increases chances of getting your changes into fastutil drastically)

4) Just look at the git diff and remove whitespace changes. It is annoying manual labor.

Speiger:
- 1: I highly suggest not to implement legacy support ways.
- 3: I rather wait on the changes to be applied. Until its finished the progress is halted.
- 4: Could you please provide examples? I am asking because i am not sure what to search for. And in worst case I am doing the inverse.

drv/ArrayFIFOQueue.drv
```diff
@@ -154,13 +171,21 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
...
 	}
 
-
 	@Override
 	public KEY_GENERIC_TYPE LAST() {
 		if (start == end) throw new NoSuchElementException();
 		return array[(end == 0 ? length : end) - 1];
 	}
-
+
```
incaseoftrouble, line 179: Whitespace example

drv/ArrayFIFOQueue.drv
```diff
@@ -192,13 +221,55 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		length = (end = size) + 1;
 		array = newArray;
 	}
-
+
```
incaseoftrouble, line 224: Whitespace example

drv/ArrayFIFOQueue.drv
```diff
@@ -192,13 +221,55 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
...
+			input = (KEY_GENERIC_TYPE[])new Object[length];
+#endif
+		}
+		if (start <= end) {
+			System.arraycopy(array, start, input, 0, end - start);
+		}
+		else {
+			System.arraycopy(array, start, input, 0, length - start);
+			System.arraycopy(array, 0, input, length - start, end);
+		}
```
incaseoftrouble, line 269: Inconsistent brace style example (I think fastutil follows the usual approach of keeping opening brace on the same line)

test/it/unimi/dsi/fastutil/ints/IntArrayFIFOQueueTest.java
```diff
@@ -158,7 +194,7 @@ public void testDeque() {
 		assertEquals(1, q.dequeueInt());
 		assertEquals(0, q.dequeueInt());
 	}
-
+
```
incaseoftrouble, line 197: Whitespace example

incaseoftrouble:
* what do you mean by "legacy support ways"? I'd rather have a documentation explaining how to properly do things than guide people indirectly.
 * Yes, that's what I'm saying. If you leave it as one big blob these changes might not get added in the near future. By splitting it up in independent parts its much easier to review and accept :-) For example, moving the remIf name, trim, and addTo changes to separate MRs, each of them can be discussed independently. Scientists notoriously have little time ;-)
 * Added some comments. But simply look at the git diff. Some changes touch regions which they shouldn't, have different style, reorder imports etc.

drv/ArrayFIFOQueue.drv
Speiger: Ok, this one i am not sure what you mean because this is the only example that clashes with your other ones.
What you mean is tabs between lines so that the curser does go back if to the beginning of the line if no code is written.
Yeah i can fix these pretty easily, but that is something I would have no guessed since you said "whitespace"

Unless you mean new lines and not whitespaces/tabs.

drv/ArrayFIFOQueue.drv
Speiger: Ok, will be adjusted.

Speiger:
- Collection.sort is just a legacy support for Java7 code that runs on Java8, thats why I highly do not suggest implementing legacy support ways.
- Honestly, I have no answer to that that does not help the conversation. So i am skipping that one.
- Yeah I kinda get what you mean but there are some inconsistencies of what you mean since you are targeting 2 things at once. But what i understood will be changed/patched.

incaseoftrouble:
* That's a different point. My argument is: `trim()` is not part of the Java `List` API despite `ArrayList` having a `trimToSize()`. If you need `trim()` functionality it is perfectly fine to use `ArrayList` instead of the generic `List` -> I think we should adhere to this convention.
 * What I am suggesting is to have one MR for "add `trim()` to the public interface", one for "add an `addTo` function" etc. These are independent things which are mixed up.
 * I am deliberately addressing multiple points. However, all of them fall under the category "codestyle" :-)

drv/ArrayFIFOQueue.drv
incaseoftrouble: Whitespace refers to tabs, spaces, newlines, etc.

Speiger:
@incaseoftrouble do you mean whitespaces in other words " " or do you mean tabs "    " or do you mean newlines?
" 
"
Or do you mean all three because right now you suggest me to run a find and replace and remove all new lines that have no line inbetween. Which would be through all classes.
If you don't mean newlines then your first example (the line you directly commented on) makes no sense.

incaseoftrouble:
Whitespace includes, among others, spaces, tabs, and newlines, see, e.g. https://en.wikipedia.org/wiki/Whitespace_character . Fixing code style consistency includes, among others, using the same line endings (LF vs CRLF), consistently using the same indentation (level and tabs vs spaces), same brace style (K&R vs 1TBS etc.) That's what I mean with "whitespace changes".

Speiger:
Ok so what you suggest me is not FastUtil conform itself. Because in every class i found (in drv files) there is newlines between functions and other things. Sometimes more sometimes less. Very confusing. I wish that drv files support a Formatting file or CodeCleaning file like java files have it. Dunno if there is a extension support.

But what i have implemented (- tabs in empty lines) is like fastutil does it. You ask me to remove empty lines which wasn't the format in the first place.

incaseoftrouble:
> Ok so what you suggest me is not FastUtil conform itself.

No =) Just look at the diff. There are, for example, additional tab characters where there should be none (e.g. AbstractMap line 70).

Speiger:
You do know there are git property settings that you can submit that would automatically enforce this rule by converting it on commit even if you wanted to submit it differently?

incaseoftrouble:
What you (probably) are talking about is `core.autocrlf` and that's something quite different (it has nothing to do with tabs, if git would magically alter tabs there would be a lot of chaos)

Speiger:
so question then because i have no idea what the exact result should be:
Tabs or whitespaces or whatever?

Speiger:
And in what scenario.

incaseoftrouble:
You simply should 1) not change lines which have nothing to do with yout functionalities and 2) the lines you commit should have the same style as the rest of the library. Don't change existing code, only adapt your code to the style of the surrounding code. I don't understand where 

For example: AbstractList#213 AbstractMap#70 ArrayFifoQueue#224 etc. should be reverted, ArrayFifoQueue#183 should be on the previous line (the rest seems to use 1TBS not K&R as brace style).

Speiger:
> There are, for example, additional tab characters where there should be none (e.g. AbstractMap line 70).

Ok that makes sense since it is 1 to much of a tab, and https://github.com/vigna/fastutil/pull/158#discussion_r509070883 this makes also sense since empty lines should not have tabs.

I get that but then i get a report like this: https://github.com/vigna/fastutil/pull/158#discussion_r509070824 where you tell me that EMPTY LINES are not allowed at all. That makes no sense at all. So i want clearifications:

Should empty lines NOT EXIST, or should empty lines STAY EMPTY (Including spaces/tabs) and you weren't able to clearly define that @incaseoftrouble

incaseoftrouble:
I don't get it. I never said anything about empty lines not being aren't allowed. ArrayFifoQueue#179 also is an added tab just as the other examples, there simply is a diff line which shouldn't be there.

Speiger:
It has no tab or space it is an empty line nothing, and you say its a whitespace example that should be deleted. That means to me empty lines are not allowed.
If it was a double empty line i would get it. But it doesn't

incaseoftrouble:
There is a tab. Github's diff viewer apparently removes trailing tabs. Open the file [directly](https://github.com/vigna/fastutil/blob/883681c643141c41e13cd8ff7634b63106a5653c/drv/ArrayFIFOQueue.drv) to see the tab

Fastutil's tab style isn't too consistent anyway. My point is that there is a diff which doesn't need to be there.

Speiger:
Ok that makes a lot of sense. 
I just tested around. I am not sure how viable that is but you could technically have a custom code Formatter and code cleaner and apply all the formatting in there and just have .drv files treated like Java files and the Formatter (if it does not break anything) could help enforce rules. 

Also text becomes a lot easier to read because all text is no longer gray. (I am going to play with that a bit and see how the result ends. (Lets hope not to bad)

drv/Collection.drv
Speiger: image
Yeah works like a charm.

Speiger:
incaseoftrouble I am sorry for my aggressiveness.

drv/Collection.drv
incaseoftrouble: Yes, I know this doesn't work :-) `data.removeIf((int t) -> t == 0);` does.

Speiger: image
Yeah works wonderfully with more optimized lambdas.

incaseoftrouble: `data.removeIf((IntPredicate) this::removeIf)` works.

My Point being: Do you have to break the public API in order to save a few characters which your IDE can automatically insert?

Speiger: So you suggest implementing functions that causes compiler warnings is a good idea in general? Because that is what you are suggesting. I mean your solution works, but the point of the change is to remove the casting requirement in general. Else why not change rem function or removeInt to remove because that's exactly the same thing.

drv/Lists.drv
Speiger: How about we implement non random access list in the first place?
A warning may be a solution but then again why not implement it into the interface and then in to the list implementation?
Could you please decide?

drv/Collection.drv
Speiger: Honestly @vigna should decide on this one.

incaseoftrouble: Its not a compiler warning, but a compiler error. And no, its far from being the same thing.

And yes of course he should decide, on all of these. I'm just trying to minimize his effort by giving you pointers for things which are very likely to cause problems. For example, changing public API for a small patch. A lot of people use this library, you can't just rename API methods without a lot of consideration. Yes it would be good if all of these methods are free of ambiguities but that is not something to be done in a minor patch.

drv/Lists.drv
incaseoftrouble: I don't see what implementing non-random access lists has to do with this discussion.

In case you want an in-place swap, I suggest using Java's `Collection#reverse` as starter. They have a good implementation which also works in linear time for properly implemented non-random access lists (instead of quadratic as this implementation would). However, fastutil doesn't use the `RandomAccess` markers I think. So it should be noted in the javadoc that this method assumes the given list to be random access.

I don't think it is sensible to add this to the list interface. I don't see any significant benefit.

drv/Collection.drv
vigna: Yeah, totally. Changing the public API is a no-no.

Speiger:
@vigna this pull request gets split into separate pull requests dedicated to each thing. Also not leaving its own classes.
So any no goes or change requests from here will be implemented.
Also #168 is finished and has tests and is only limited to priority queue and its implementations.

Speiger:
@vigna @incaseoftrouble Lists have now the dedicated changes moved to #169 
All discussion points were accounted for.

Speiger:
Also @vigna should i put the tutorial directly into the wiki? Because this technically is not needed to be part of the sourcecode itself?
And if you approve the tutorial, then this can be done right now. (You left Wiki section open for everyone to edit)

Issue #101, opened 2018-06-15
Various default methods from java.util.Map don't work correctly if defRetValue is not null


KidsDontPlay:
Version: fastutil-8.2.1
```java
Object2ObjectOpenHashMap<String, String> map = new Object2ObjectOpenHashMap<String, String>();
map.defaultReturnValue("undefined");

System.out.println(map.getOrDefault("bar", "nak")); //prints 'undefined' instead of 'nak'
	
map.putIfAbsent("foo", "one"); // 'foo=>one' is not added, see next line
System.out.println(map.get("foo")); //prints 'undefined' instead of 'one'
	
System.out.println(map.replace("bar", "two")); //prints 'undefined' instead of 'null' and 'bar=>two' is added (should not be) 
	
System.out.println(map.computeIfAbsent("qux", s -> "three")); //prints 'undefined instead of 'three' and 'qux=>three is not added (should be)
	
System.out.println(map.computeIfPresent("def", (a, b) -> "four")); //prints 'four' instead of null and 'def=>four is added (should not be)
```

vigna:
Ouch, I can see where we botched. We let Java's default implementation go through in the Object2Object case, but they don't take into account default values (which are a fastutil feature). Java's default implementation sees a non-null return value from the underlying map and returns that value.

I have to think what to do here. We might want just to document this behaviour.

develar:
It is also performance issue. `computeIfAbsent` in java hashmap is optimized and doesn't calculate hash twice (the same reason why there is `addOrGet` method for int set).

vigna:
This should be fixed by 82e2ad359fd09fbe8dfb7ef40a0fc5a6abb60976.

PR #162, opened 2020-07-05
Fast in-place compute/merge methods in OpenHashMap classes

gregsh:
Some things I'd like to highlight:

* `null` handling is preserved for default Map methods, so no `defaultValue` support (checked in tests)
* `0` is treated as `null` element in `Double/Float2xxx` maps which seems rather suspicious
* `xxxMap#computeIfAbsent` methods with `xxxUnaryOperator` can be named `computeIfAbsentFast` to avoid clashing

### DIFF
--- a/drv/OpenHashMap.drv
+++ b/drv/OpenHashMap.drv
@@ -1260,6 +1260,74 @@ public class OPEN_HASH_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENE
 		return value[pos] = VALUE_CLASS2TYPE(newValue);
 	}
 
+#if VALUES_INT_LONG_DOUBLE
+
+#if KEYS_PRIMITIVE && KEYS_AND_VALUES_WIDENED_ARE_SAME
+	public VALUE_GENERIC_TYPE COMPUTE_IF_PRESENT_PRIMITIVE(final KEY_GENERIC_TYPE k, final JDK_PRIMITIVE_VALUE_BINARY_OPERATOR remappingFunction) {
+		java.util.Objects.requireNonNull(remappingFunction);
+		final int pos = find(k);
+		if (pos < 0) return defRetValue;
+		final VALUE_GENERIC_TYPE newValue = VALUE_NARROWING(remappingFunction.JDK_PRIMITIVE_VALUE_OPERATOR_APPLY(k, value[pos]));
+		if (newValue == defRetValue) {
+			if (KEY_EQUALS_NULL(k)) removeNullEntry();
+			else removeEntry(pos);
+			return defRetValue;
+		}
+
+		return value[pos] = newValue;
+	}
+
+	public VALUE_GENERIC_TYPE COMPUTE_PRIMITIVE(final KEY_GENERIC_TYPE k, final JDK_PRIMITIVE_VALUE_BINARY_OPERATOR remappingFunction) {
+		java.util.Objects.requireNonNull(remappingFunction);
+		final int pos = find(k);
+		VALUE_GENERIC_TYPE newVal = VALUE_NARROWING(remappingFunction.JDK_PRIMITIVE_VALUE_OPERATOR_APPLY(k, pos >= 0 ? value[pos] : defRetValue));
+		if (newVal == defRetValue) {
+			if (pos >= 0) {
+				if (KEY_EQUALS_NULL(k)) removeNullEntry();
+				else removeEntry(pos);
+			}
+			return defRetValue;
+		}
+
+		if (pos < 0) {
+			insert(-pos - 1, k, newVal);
+			return newVal;
+		}
+
+		return value[pos] = newVal;
+	}
+#endif
+
+	public VALUE_GENERIC_TYPE MERGE_PRIMITIVE(final KEY_GENERIC_TYPE k, final VALUE_GENERIC_TYPE v, final JDK_PRIMITIVE_VALUE_BINARY_OPERATOR remappingFunction) {
+		java.util.Objects.requireNonNull(remappingFunction);
+
+		final int pos = find(k);
+		if (pos < 0 || value[pos] == defRetValue) {
+			if (v == defRetValue) return defRetValue;
+
+			insert(-pos - 1, k, v);
+			return v;
+		}
+		VALUE_GENERIC_TYPE newVal = VALUE_NARROWING(remappingFunction.JDK_PRIMITIVE_VALUE_OPERATOR_APPLY(value[pos], v));
+		if (newVal == defRetValue) {
+			if (pos >= 0) {
+				if (KEY_EQUALS_NULL(k)) removeNullEntry();
+				else removeEntry(pos);
+			}
+			return defRetValue;
+		}
+
+		return value[pos] = newVal;
+	}
+
+#if KEYS_REFERENCE
+	public VALUE_GENERIC_TYPE MERGE_TYPE_SPECIFIC(final KEY_GENERIC_TYPE k, final VALUE_GENERIC_TYPE v, final VALUE_PACKAGE.VALUE_BINARY_OPERATOR remappingFunction) {
+		java.util.Objects.requireNonNull(remappingFunction);
+		return MERGE_PRIMITIVE(k, v, (JDK_PRIMITIVE_VALUE_BINARY_OPERATOR)remappingFunction);
+	}
+#endif
+
+#endif
 
 #endif
 
--- a/drv/OpenHashMap.drv
+++ b/drv/OpenHashMap.drv
@@ -1263,6 +1263,108 @@ public class OPEN_HASH_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENE
 
 #endif
 
+
+	/** {@inheritDoc} */
+	@Override
+	SUPPRESS_WARNINGS_KEY_UNCHECKED
+	public VALUE_GENERIC_CLASS getOrDefault(final Object k, final VALUE_GENERIC_CLASS defaultValue) {
+		if (k == null) return defaultValue;
+		if (KEY_EQUALS_NULL(KEY_CLASS_CAST k)) return containsNullKey ? value[n] : defaultValue;
+
+		KEY_GENERIC_CLASS curr;
+		final KEY_GENERIC_TYPE[] key = this.key;
+		int pos;
+
+		// The starting point.
+		if (KEY_IS_NULL(curr = key[pos = KEY2INTHASH_CAST(KEY_CLASS_CAST k) & mask])) return defaultValue;
+		if (KEY_EQUALS_NOT_NULL_CAST(KEY_CLASS_CAST k, curr)) return value[pos];
+		// There's always an unused entry.
+		while(true) {
+			if (KEY_IS_NULL(curr = key[pos = (pos + 1) & mask])) return defaultValue;
+			if (KEY_EQUALS_NOT_NULL_CAST(KEY_CLASS_CAST k, curr)) return value[pos];
+		}
+	}
+
+	/** {@inheritDoc} */
+	@Override
+	public VALUE_GENERIC_CLASS computeIfAbsent(final KEY_GENERIC_CLASS k, final java.util.function.Function<? super KEY_GENERIC_CLASS, ? extends VALUE_GENERIC_CLASS> mappingFunction) {
+		java.util.Objects.requireNonNull(mappingFunction);
+		if (k == null) return null;
+		final int pos = find(k);
+		if (pos >= 0) return value[pos];
+		final VALUE_GENERIC_CLASS newValue = mappingFunction.apply(k);
+		if (newValue == null) return null;
+		final VALUE_GENERIC_CLASS v = VALUE_CLASS2TYPE(newValue);
+		insert(-pos - 1, k, v);
+		return v;
+	}
+
+	/** {@inheritDoc} */
+	@Override
+	public VALUE_GENERIC_CLASS computeIfPresent(final KEY_GENERIC_CLASS k, final java.util.function.BiFunction<? super KEY_GENERIC_CLASS, ? super VALUE_GENERIC_CLASS, ? extends VALUE_GENERIC_CLASS> remappingFunction) {
+		java.util.Objects.requireNonNull(remappingFunction);
+		if (k == null) return null;
+		final int pos = find(k);
+		if (pos < 0) return null;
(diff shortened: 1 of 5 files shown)


gregsh:
Friendly bump. Were my changes of any use?

incaseoftrouble:
One note: I think using `...AsInt` as suffix is a bit more in line with JDK convention than `...Fast`. For `computeIfAbsentFast` I'd rather say `computeAsIntIfAbsent` or `computeIfAbsentAsInt` (replacing `Int` with the respective type, of course). But that's a matter of taste. Leading to the next point.

Also, please replace the `CONCAT` with concrete names. Something like `COMPUTE_IF_ABSENT` macros etc. (which can be defiend using `CONCAT` of course). This way, other parts of the implementation can call these methods safely (for example, wiriting `this.COMPUTE_IF_ABSENT(foo, bar)`). And these names can be easily changed at a single place.

gregsh:
@incaseoftrouble thanks for response

I've extracted CONCAT'd names, and tried to replace Fast prefix with something you've proposed.

1. For maps with primitive keys (e.g. `Int2IntOpenHashMap`) everything goes smooth. `fastutil` already uses names like `mergeInt` in maps with reference keys, so I switched to just `mergeInt`, `computeIntIfPresent` (not `mergeAsInt`, etc.).
2. For maps with reference keys (e.g. ``) we have a problem. They already use `computeInt` and `mergeInt` names for methods with `java.util.function.BiFunction` argument.
  2.1 We can't use `Int` suffix due to lambda-unfriendly overload
  2.2 We can't use `AsInt` suffix because `mergeInt` and `mergeAsInt` or `mergeIntAsInt` methods in the same class look indistinguishable.

Also "fast" methods in maps with primitive and non-primitive keys should have similar names, shouldn't they?
I'm not sure how to deal with all that without breaking the current API.

incaseoftrouble:
Ugh. What a headache.

What if instead of providing a new method we check whether the given remapping function (in `COMPUTE_IF_PRESENT`) is `instanceof JDK_PRIMITIVE_BI_FUNCTION`? Unfortunately, these methods aren't bulk operations, so the instanceof is called quite often. But I guess this is better than providing yet another specialized method. The difference is that this does not cause boxing / unboxing, so the only overhead is a cast, no objects are being allocated. Also, the instanceof check can actually be delayed until the function needs to be called.

gregsh:
On the second thought, that lambda cast may be the lesser evil, if we don't want to add new names.
I've chosen that in my last commit, check out the tests.

gregsh:
On the second thought the simple `mergeInt` overload looks not that bad, consider:
`mergeInt(1, 2, (int a, int b) -> a)` or `m.mergeInt(2, 2, (IntBinaryOperator)(a, b) -> a)`.

Source-compatibility looks to be broken, but binary compatibility would stay intact.
Existing code would need explicit types to be compiled, e.g. `mergeInt(1, 2, (Integer a, Integer b) -> a)`.

What do you think?

incaseoftrouble:
Well, source compatibility is important. I'd rather have different names than adding that cast (pragmatically, that's less characters). Just compare: `map.mergeIntInt(1, 2, Integer::sum)` to `map.mergeInt(1, 2, (IntBinaryOperator) Integer::sum)`. I personally prefer the former.

But its maybe for @vigna to decide :-)

vigna:
Welcome to naming headaches in fastutil—one of my favorite pastimes in the last 22 years 😂.

So, I examined the code and it's great work. Sorry for the delay in answering. Let's try to work out the naming problems and merge this in.

A good question: why not creating type-specific BiFunctions? From what I understand in that case here the advantage is only in the return type...

incaseoftrouble:
> A good question: why not creating type-specific BiFunctions? From what I understand in that case here the advantage is only in the return type...

Well, if you want all then you'd cubic amount of functions (x, y -> z), maybe we can generate one for a) the "primary" types (int, long, double; for computations it doesn't matter anyway) and b) only where Java doesn't have one already.

vigna:
After having see @techsy730 clever idea for forEachRemaining(), I think we can try to do the same here. The idea is to have a type-specific BinaryOperator class, and implement that. Then we implement the version based on the JDK primitive thing, if it exists, and we add a method accepting a type-specific BinaryOperator. The trick is that the latter is more specific than _both_ the other options, and so it is automatically chosen by the lambda.

The only difference is a little code duplication—in the face of Iterator, @techsy730 added a stub delegating the method. But in this case is a single delegation for a batch of work. In this cases I would simply duplicate the (short) code for the two versions for int, long and double. In the other cases there's no duplication.

gregsh:
@vigna Do I understand correctly that you are referring to the following solution?

```
import java.util.function.*;

class Scratch {
  public static void func(ObjectIntMap<Integer> map) {
    map.mergeInt(10, 20, (a, b) -> a + b);  // ok, IntBinaryOp inferred
    map.mergeInt(10, 20, Integer::sum);  // ok, IntBinaryOp inferred

    // We have preserved source and binary compatibility but:
    // 1. IntBinaryOp interface and the related method are here forever for binary compatibility
    // 2. the following code is now also possible 

    map.mergeInt(10, 20, new IntBinaryOp() {
      @Override public int applyAsInt(int left, int right) { return left + right; } // <--- that one is used
      @Override public Integer apply(Integer left, Integer right) { return left - right; } // ignored
    });
  }

  interface ObjectIntMap<K> {
    int mergeInt(K key, int v, BiFunction<? super Integer, ? super Integer, ? extends Integer> func); // JDK boxed
    int mergeInt(K k, int v, IntBinaryOperator func); // JDK primitive

    // more specific version with out custom interface
    int mergeInt(K k, int v, IntBinaryOp func); // duplicate JDK primitive implementation
  }

  @FunctionalInterface
  interface IntBinaryOp extends IntBinaryOperator, BiFunction<Integer, Integer, Integer> {
    @Override
    default Integer apply(Integer integer, Integer integer2) {
      return this.applyAsInt(integer.intValue(), integer2.intValue());
    }
  }
}
```

vigna:
Almost. IntBinaryOp has an apply(int, int) metho,which is the one delegating to applyAsInt() is available. As usual, the method inherited from BinaryOperator delegates to the type-specific one and is deprecated. The ignored method would give a warning as you'd be overriding a deprecated method. Another advantage is that we would have composition for all types. 

It is also possible in the future that we do something like @techsy730 work on streams, in which we also adapt JDK primitive operators (e.g., you can use an IntBinaryOperator with character values).

vigna:
So, I slightly hijacked this PR in the sense that I implemented type-specific binary operators. The modifications to gencsources.sh are no longer necessary. Could you try to bring this PR in the form we discussed above?

gregsh:
I've pushed changes for Int/Long/Double maps as we discussed above (had to fix BinaryOperator.drv a bit).
We can use new type-specific operators for all other primitive maps.

drv/BinaryOperator.drv
```diff
@@ -60,6 +60,10 @@ public interface KEY_BINARY_OPERATOR KEY_GENERIC extends BinaryOperator<KEY_GENE
 	@Override
 	@SuppressWarnings("boxing")
 	default KEY_CLASS apply(final KEY_CLASS x, final KEY_CLASS y) {
+#if !defined JDK_PRIMITIVE_BINARY_OPERATOR || KEY_WIDENED
```
vigna, line 63: I do not understand this modification. The only method of the functional interface is apply() with primitive types. The method you are delegating to delegates to that method.

vigna:
OK, I understand a bit better now the situation. The problem is that this has to be fixed first at the _interface_ level, or open hash maps will have type-specific method no other implementation has.

If I understand correctly, we would add compute() methods only if the type of keys and values is the same (as otherwise bifunctions would be too many) and merge for all types because in that case the bifunction it is just a binary operator. Presently, moreover, the only code is for the JDK special types int/long/double, for which there are functions. We would get some ambiguity when there is an object-based operator and a JDK operator.

drv/BinaryOperator.drv
gregsh: I was fixing the following:
```
public interface IntBinaryOperator extends BinaryOperator<Integer>, java.util.function.IntBinaryOperator {
	  default Integer apply(final Integer x, final Integer y) {
// it was a recursive apply(Integer, Integer) call here before the change
// as apply(int, int) is not generated for IntBinaryOperator
	 return applyAsInt(x.intValue(), y.intValue()); 
	}
}
```

gregsh:
> OK, I understand a bit better now the situation. The problem is that this has to be fixed first at the interface level, or open hash maps will have type-specific method no other implementation has.

Sorry, I was referring to Byte/Char/Short/.. **open** hash maps, as those types now have their binary operators missing in JDK thanks to `BinaryOperator.drv`. Or we can use widened operators there for JDK methods. I don't care about them much, so it is for you to decide. I.e.:

```
  interface Object2CharOpenHashMap<K> {
    int mergeChar(K k, char v, BiFunction<? super Character, ? super Character, ? extends Character> func); // existing
 // int mergeChar(K k, char v, IntBinaryOperator func); // optional JDK primitive for widened type
    int mergeChar(K k, char v, CharacterBinaryOp func); // type-specific operator
  }
```

> If I understand correctly, we would add compute() ...

Yes, `compute` and `computeIfPresent` have same signature `(K, (K, V)-> V) -> V`, so type specific methods with `BinaryOperator` for `(K, V)->V` argument is generated only when K and V have equal wide type.

vigna:
Hey, we're working for a better world—in the sense of a more readable #ifdef'ing of the code. We'll come back to this PR as soon as things are stabilized (hopefully soon).

drv/BinaryOperator.drv
vigna: I see. No, it's my code that's wrong, in the sense that I did not realize JDK's operators have that funny name. I would like to follow the same structure as in other classes as Function, so there will be an apply() method and applyAsInt() will delegate.

vigna:
So, there has been a lot of internal reworking inside fastutil, so we will need to bring this up to date with those reworking. I'm also thinking about including functions K⨉K->V and K⨉V->V, which are relatively limited in number, so we can have type-specific implementation of these functions (compute, merge, etc.) for all types. We're very close to a release, so I would wait for the next one (soon) to include this stuff and possibly solve #101 , which is strictly related.

gregsh:
Something wrong with 4be250e92d8ce6eb0efcaa4545046067340aaa9b. MERGE def is deleted and a new method with MERGE name is added in `Map.drv`. I can't compile my PR because it uses MERGE=merge and I get duplicate `merge` methods in error in `Int2ObjectMap` and etc.

vigna:
Urgh. I made so many changes—some mistakes slipped through. Thank you for reporting this, I'm looking into it...

vigna:
That should be fixed with 6bacc8990. There were several cases in which the method was compiled in, but it wasn't necessary.

I think I didn't left out any case (or there would be pre-existing methods that do not exist anymore).

gregsh:
Compiles OK now. Are there any questions left for this PR?

vigna:
Almost there.

- To bring this in line with the overall restyling, we need to avoid #ifdef. There are several things you can use, KEYS_INT_LONG_DOUBLE, etc., but we have tried to eliminate anything that is not readable immediately. `#ifdef SOME_METHOD` makes incredibly difficult for the reader to understand when the code will be compiled in.
- Did you try whether no ambiguity arises? You should be able to write lambdas for all these methods without any type indications.
- I would like to add functions XY->Y and XX->Y to complete the picture.

So the plan is to release 8.5.0 (which certainly will have bug reports), and while it stabilizes complete this PR. I would like to have those functions in because there might be ambiguity problems we do not realize immediately. Then we can have default methods in Map.drv, and these optimized methods for hash maps.

Adding the functions is a bit tricky, because to preserve your idea, which I think is very good (always supporting the JDK binary operator in some way) we need some conditional compilation similar to that for functions. Like, CharChar2ByteFunction should extend IntBinaryOperator. In this way we can play that "three method game" that we played with all other methods (one method has a type that extends both the other two).

Does that sound sensible?

gregsh:
> To bring this in line with the overall restyling, we need to avoid #ifdef.

Do you have some specific code lines in mind?

> Did you try whether no ambiguity arises?

Sure. `Object2IntOpenHashMapTest.java` has new tests for that.

> I would like to add functions XY->Y and XX->Y to complete the picture.

We can add that later, while already testing the new API...

vigna:
> Do you have some specific code lines in mind?

Everything that is not explicit. `#ifdef COMPUTE`, for example. In general, no `#ifdef` (stuff like `KEYS_INT_LONG_DOUBLE` has an actual value 0 or 1, so you can #if it).
 
> Sure. `Object2IntOpenHashMapTest.java` has new tests for that.
> 
> > I would like to add functions XY->Y and XX->Y to complete the picture.
> 
> We can add that later, while already testing the new API...

Mmmh. My only fear is that adding methods we might discover that some change of signature is necessary, and then we cannot rollback. Let me think, I might add the functions now, it's not that much work.

gregsh:
> To bring this in line with the overall restyling, we need to avoid #ifdef.

Done. Anything else?

vigna:
If I understand correcty, we have here two type of methods: more efficient implementations of existing interface methods, and some completely new methods (`*_PRIMITIVE`) that are not present in the type-specific Map interface. Am I right?

gregsh:
> If I understand correcty, we have here two type of methods: more efficient implementations of existing interface methods, and some completely new methods (*_PRIMITIVE) that are not present in the type-specific Map interface. Am I right?

Right

vigna:
I think it should be easy to lift those method to default interface implementations no?

vigna:
The rationale here is that probably the methods are fine and there is no need to include hundreds of functions to make them align with the strategy used elsewhere, but it is not a good idea to have methods that appear just in an implementation as people might not get access to them. From what I see, the four `_PRIMITIVE` methods's code can be very easily lifted up to Map.drv with small changes, and then we would have an backward-compatible interface extension and an optimized implementation for hash-based maps.

vigna:
Did I miss something?

gregsh:
I guess no, I just wait on your decision on this PR regarding `OpenHashMap`.

Lifting some methods to `Map.drv` seems to be a separate task.
Potential name conflicts could be easily tested at the current state of things. 
No real implementations, additional tests, docs, etc. are required.

vigna:
I waiting for an answer to a question I posed three weeks ago. I guess I'll wait more. 🤷🏻‍♂️

gregsh:
@vigna I don't understand. Did I miss a question?

vigna:
82e2ad359fd09fbe8dfb7ef40a0fc5a6abb60976 has added some interface methods.

PR #164, opened 2020-08-19
Avoid allocating large arrays that crash the JVM

electrum:
Allocating an array of size close to 2^31 will cause an internal `OutOfMemoryError: Requested array size exceeds VM limit`
that triggers any configured JVM OOM behavior such as `-XX:+ExitOnOutOfMemoryError` or `-XX:+CrashOnOutOfMemoryError`. These errors can be difficult to debug. Instead, we can throw a normal `OutOfMemoryError` exception.

### DIFF
--- a/drv/Arrays.drv
+++ b/drv/Arrays.drv
@@ -174,6 +174,10 @@ public final class ARRAYS {
 	 * entries are the same as those of {@code array}.
 	 */
 	public static KEY_GENERIC KEY_GENERIC_TYPE[] forceCapacity(final KEY_GENERIC_TYPE[] array, final int length, final int preserve) {
+		if (length > Arrays.MAX_ARRAY_SIZE) {
+			throw new OutOfMemoryError("Requested array size " + length + " exceeds the maximum size of " + Arrays.MAX_ARRAY_SIZE);
+		}
+
 		final KEY_GENERIC_TYPE t[] =
 #if KEY_CLASS_Object
 			newArray(array, length);


vigna:
This is not a bad idea but the constant is already in it.unimi.dsi.fastutil.Arrays. I'm also not entirely sure that if the user really want to force their size we should get in the way. One thing is grow(), in which we choose the capacity. forceCapacity() is another matter.

electrum:
Thanks for pointing out the existing constant. Note that we aren't getting in the users way, as the allocation would fail anyway. This simply prevents the JVM from crashing.

I added this check in `forceCapacity()` since that is the place where the allocation happens. Would you be open to adding this check in the private `grow()` method?

As a specific example, we would like to prevent this type of call sequence from crashing the JVM:
```java
LongArrayList list = new LongArrayList();
for (int i = 0; i < Integer.MAX_VALUE; i++) {
    list.add(i);
}

**Comment by vigna** [comment]
Well, but it's already checked in grow()...

**Comment by electrum** [comment]
The problem is this line of code in `grow()`:
```java
capacity = (int) Math.max(
    Math.min((long) a.length + (a.length >> 1), it.unimi.dsi.fastutil.Arrays.MAX_ARRAY_SIZE), capacity);
```
The `max()` will choose `capacity` if it is larger than `MAX_ARRAY_SIZE`.

I was able to reproduce the behavior on my laptop with this variant of the above (requires less memory):
```java
ByteArrayList list = new ByteArrayList();
for (int i = 0; i < Integer.MAX_VALUE; i++) {
    list.add((byte) 0);
}
```
It fails with this: `java.lang.OutOfMemoryError: Requested array size exceeds VM limit`

Setting a debugger breakpoint for `OutOfMemoryError`, I can see that `forceCapacity()` is trying to allocate a `byte[]` of size `2147483646`, which is larger than the JVM allows.

vigna:
Ah OK now I see your point. You'd do first the max and then the min, right?

electrum:
That would fix it, though it might violate the contract of the method, if the expectation is that the array will be at least as big as the `capacity` argument when it returns. It has to throw somewhere for sizes in the small invalid range near `MAX_VALUE`. The JVM will throw (or crash) anyway if we don’t.

Are you concerned with allowing sizes that are larger than `MAX_VALUE - 8` if the JVM would allow them? In my testing above, the actual cutoff was `MAX_VALUE - 2`, but I don’t think that matters in practice.

sirinath:
I think the best bet is to stick and adjustment which:
- accounts for object headers
- accounts for `length` field
- accounts for 2GB boundary

vigna:
Actually, let's all have a look at JDK's ArraysSupport.newLength()—I think the approach there is the best.

electrum:
I updated the PR to use the existing constant, which is the same as the one used in the JDK. This will prevent the JVM from crashing, with no practical difference in behavior.

vigna:
My point is that the Java JDK tries to not *grow* an array more than ARRAY_SIZE, but does not prevent the user from explicitly requesting that size. I think this is essential—you cannot cripple the library by making impossible for a user to allocate an array of size Integer.MAX_VALUE if his/her JVM allows it.

PR #168, opened 2020-10-23
Priority queue improvements

Speiger:
A more deep improvement of PriorityQueues.
All new functions are implemented in all PriorityQueue implementations.

Also as suggested from incaseoftrouble i made this a dedicated patch.
So it gets implemented much more quickly instead of waiting longer.
Other parts of the patch will get transferred to dedicated patches.

Table of Contents:
- Added: Peek function for looking ahead on values.
- Added: trim function to interface.
- Added: (FIFOOnly) Constructor that supports arrays for loading/priming/wrapping.
- Added: toArray function. So data can be cloned without emptying the queue.
- Changed: (FIFOOnly) expand/shrink are made protected because they are usually the target of changes to prevent them on executing. Reduces the amount of overrides required.
- Added: Remove function so values can be removed out of the PriorityQueues. (Removes only one value)

Edit: Improvements to description
Note: Pull request is considered ready for review from my side.

### DIFF
--- a/drv/PriorityQueue.drv
+++ b/drv/PriorityQueue.drv
@@ -59,6 +59,23 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	 */
 
 	default KEY_GENERIC_TYPE LAST() { throw new UnsupportedOperationException(); }
+	
+	/** Returns the index element of the queue.
+	 * 
+ 	 * <p>This default implementation just throws an {@link UnsupportedOperationException}.
+ 	 * @see #peek()
+	 * @param index of the element.
+	 * @return the indexed element.
+	 * @throws NoSuchElementException if the index is not in the range
+	 */
+	
+	default KEY_GENERIC_TYPE PEEK(int index) { throw new UnsupportedOperationException(); }
+	
+	boolean REMOVE_KEY(KEY_GENERIC_TYPE x);
+	
+	KEY_GENERIC_TYPE[] TO_KEY_ARRAY();
+	
+	KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input);
 
 	/** Returns the comparator associated with this priority queue, or null if it uses its elements' natural ordering.
 	 *
@@ -68,7 +85,7 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	 */
 	@Override
 	KEY_COMPARATOR comparator();
-
+	
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
@@ -96,4 +113,32 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	@Deprecated
 	@Override
 	default KEY_GENERIC_CLASS last() { return KEY2OBJ(LAST()); }
+	
+	/** {@inheritDoc}
+	 * <p>This default implementation delegates to the corresponding type-specific method.
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
+	default KEY_GENERIC_CLASS peek(int index) { return KEY2OBJ(PEEK(index)); }
+	
+	/** {@inheritDoc}
+	 * <p>This default implementation delegates to the corresponding type-specific method.
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
+	default boolean remove(KEY_GENERIC_CLASS x) { return remove(x.KEY_VALUE()); }
+	
+	/** {@inheritDoc}
+	 * <p>This default implementation delegates to the corresponding type-specific method.
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
+	default KEY_GENERIC_CLASS[] toArray() { return toArray((KEY_GENERIC_CLASS[])null);}
+	
+	/** {@inheritDoc}
+	 * <p>This default implementation delegates to the corresponding type-specific method.
+	 * @deprecated Please use the corresponding type-specific method instead. */
+	@Deprecated
+	@Override
+	default KEY_GENERIC_CLASS[] toArray(KEY_GENERIC_CLASS[] input) { throw new UnsupportedOperationException(); }
 }
--- a/drv/PriorityQueue.drv
+++ b/drv/PriorityQueue.drv
@@ -59,7 +59,7 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	 */
 
 	default KEY_GENERIC_TYPE LAST() { throw new UnsupportedOperationException(); }
-	
+
 	/** Returns the index element of the queue.
 	 * 
  	 * <p>This default implementation just throws an {@link UnsupportedOperationException}.
@@ -68,13 +68,13 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	 * @return the indexed element.
 	 * @throws NoSuchElementException if the index is not in the range
 	 */
-	
+
 	default KEY_GENERIC_TYPE PEEK(int index) { throw new UnsupportedOperationException(); }
-	
+
 	boolean REMOVE_KEY(KEY_GENERIC_TYPE x);
-	
+
 	KEY_GENERIC_TYPE[] TO_KEY_ARRAY();
-	
+
 	KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input);
 
 	/** Returns the comparator associated with this priority queue, or null if it uses its elements' natural ordering.
@@ -85,7 +85,7 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	 */
 	@Override
 	KEY_COMPARATOR comparator();
-	
+
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
@@ -113,28 +113,28 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
 	@Deprecated
 	@Override
 	default KEY_GENERIC_CLASS last() { return KEY2OBJ(LAST()); }
-	
+
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
 	@Deprecated
 	@Override
 	default KEY_GENERIC_CLASS peek(int index) { return KEY2OBJ(PEEK(index)); }
-	
+
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
 	@Deprecated
 	@Override
 	default boolean remove(KEY_GENERIC_CLASS x) { return remove(x.KEY_VALUE()); }
-	
+
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
 	@Deprecated
 	@Override
 	default KEY_GENERIC_CLASS[] toArray() { return toArray((KEY_GENERIC_CLASS[])null);}
-	
+
 	/** {@inheritDoc}
 	 * <p>This default implementation delegates to the corresponding type-specific method.
 	 * @deprecated Please use the corresponding type-specific method instead. */
--- a/src/it/unimi/dsi/fastutil/PriorityQueue.java
+++ b/src/it/unimi/dsi/fastutil/PriorityQueue.java
@@ -75,7 +75,7 @@ public interface PriorityQueue<K> {
 	 */
 
 	void clear();
-	
+
 	/** Trims the underlying array so that it has exactly {@link #size()} elements. */
 	void trim();
 
@@ -103,26 +103,27 @@ public interface PriorityQueue<K> {
 	 * @return the indexed element.
 	 * @throws NoSuchElementException if the index is not in the range
 	 */
-	
+
 	default K peek(int index){ throw new UnsupportedOperationException(); }
-	
+
+	/*Removes a single instance of the specified element from this */
 	boolean remove(K value);
-	
-	
+
 	/** Notifies the queue that the {@linkplain #first() first} element has changed (optional operation).
 	 * <p>This default implementation just throws an {@link UnsupportedOperationException}. 
 	 */
 
 	default void changed() { throw new UnsupportedOperationException(); }
 
-
 	/** Returns the comparator associated with this queue, or {@code null} if it uses its elements' natural ordering.
 	 *
 	 * @return the comparator associated with this sorted set, or {@code null} if it uses its elements' natural ordering.
 	 */
 	Comparator<? super K> comparator();
-	
+
+	/** Returns an array containing all of the elements in this PriorityQueue. */
 	K[] toArray();
-	
+
+	/** Returns an array containing all of the elements in this PriorityQueue. */
 	K[] toArray(K[] input);
 }
--- a/src/it/unimi/dsi/fastutil/PriorityQueue.java
+++ b/src/it/unimi/dsi/fastutil/PriorityQueue.java
@@ -75,6 +75,9 @@ public interface PriorityQueue<K> {
 	 */
 
 	void clear();
+	
+	/** Trims the underlying array so that it has exactly {@link #size()} elements. */
+	void trim();
 
 	/** Returns the first element of the queue.
 	 *
@@ -92,7 +95,20 @@ public interface PriorityQueue<K> {
 	 */
 
 	default K last() { throw new UnsupportedOperationException(); }
-
+	
+	/** Returns the index element of the queue.
+	 * 
+ 	 * <p>This default implementation just throws an {@link UnsupportedOperationException}.
+	 * @param index of the element.
+	 * @return the indexed element.
+	 * @throws NoSuchElementException if the index is not in the range
+	 */
+	
+	default K peek(int index){ throw new UnsupportedOperationException(); }
+	
+	boolean remove(K value);
+	
+	
 	/** Notifies the queue that the {@linkplain #first() first} element has changed (optional operation).
 	 * <p>This default implementation just throws an {@link UnsupportedOperationException}. 
 	 */
@@ -105,4 +121,8 @@ public interface PriorityQueue<K> {
 	 * @return the comparator associated with this sorted set, or {@code null} if it uses its elements' natural ordering.
 	 */
 	Comparator<? super K> comparator();
+	
+	K[] toArray();
+	
+	K[] toArray(K[] input);
 }
--- a/src/it/unimi/dsi/fastutil/PriorityQueues.java
+++ b/src/it/unimi/dsi/fastutil/PriorityQueues.java
@@ -61,7 +61,7 @@ public class PriorityQueues {
 
 		@Override
 		public Object last() { throw new NoSuchElementException(); }
-		
+
 		@Override
 		public boolean remove(Object value) { return false; }
 
@@ -70,7 +70,7 @@ public class PriorityQueues {
 
 		@Override
 		public Comparator<?> comparator() { return null; }
-		
+
 		@Override
 		public void trim() { throw new NoSuchElementException(); }
 
@@ -139,7 +139,7 @@ public class PriorityQueues {
 
 		@Override
 		public K last() { synchronized(sync) { return q.last(); } }
-		
+
 		@Override
 		public K peek(int index) { synchronized(sync) { return q.peek(index); } }
 
@@ -154,6 +154,7 @@ public class PriorityQueues {
 
 		@Override
 		public void changed() { synchronized(sync) { q.changed(); } }
+
 		@Override
 		public void trim() { synchronized(sync) { q.trim(); } }
 
@@ -182,7 +183,6 @@ public class PriorityQueues {
 			synchronized(sync) { s.defaultWriteObject(); }
 		}
 	}
-
 
 	/** Returns a synchronized priority queue backed by the specified priority queue.
 	 *
--- a/src/it/unimi/dsi/fastutil/PriorityQueues.java
+++ b/src/it/unimi/dsi/fastutil/PriorityQueues.java
@@ -17,7 +17,6 @@ package it.unimi.dsi.fastutil;
  */
 
 import java.io.Serializable;
-
 import java.util.Comparator;
 import java.util.NoSuchElementException;
 
@@ -62,12 +61,27 @@ public class PriorityQueues {
 
 		@Override
 		public Object last() { throw new NoSuchElementException(); }
+		
+		@Override
+		public boolean remove(Object value) { return false; }
 
 		@Override
 		public void changed() { throw new NoSuchElementException(); }
 
 		@Override
 		public Comparator<?> comparator() { return null; }
+		
+		@Override
+		public void trim() { throw new NoSuchElementException(); }
+
+		@Override
+		public Object peek(int index) { throw new NoSuchElementException(); }
+
+		@Override
+		public Object[] toArray() { throw new NoSuchElementException(); }
+
+		@Override
+		public Object[] toArray(Object[] input) { throw new NoSuchElementException(); }
 
 		@Override
 		public Object clone() { return EMPTY_QUEUE; }
@@ -125,6 +139,9 @@ public class PriorityQueues {
 
 		@Override
 		public K last() { synchronized(sync) { return q.last(); } }
+		
+		@Override
+		public K peek(int index) { synchronized(sync) { return q.peek(index); } }
 
 		@Override
 		public boolean isEmpty() { synchronized(sync) { return q.isEmpty(); } }
@@ -137,9 +154,20 @@ public class PriorityQueues {
 
 		@Override
 		public void changed() { synchronized(sync) { q.changed(); } }
+		@Override
+		public void trim() { synchronized(sync) { q.trim(); } }
 
 		@Override
 		public Comparator <? super K> comparator() { synchronized(sync) { return q.comparator(); } }
+
+		@Override
+		public boolean remove(K value) { synchronized(sync) { return q.remove(value); } }
+
+		@Override
+		public K[] toArray() { synchronized(sync) { return q.toArray(); } }
+
+		@Override
+		public K[] toArray(K[] input) { synchronized(sync) { return q.toArray(input); } }
 
 		@Override
 		public String toString() { synchronized(sync) { return q.toString(); } }
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -25,7 +25,6 @@ import java.util.Comparator;
 import java.io.Serializable;
 import it.unimi.dsi.fastutil.HashCommon;
 import it.unimi.dsi.fastutil.PriorityQueue;
-import it.unimi.dsi.fastutil.ints.IntArrayList;
 
 import java.util.NoSuchElementException;
 
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -305,16 +305,21 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 
 #if KEYS_PRIMITIVE
 	/** Returns an array containing all of the elements in this PriorityQueue. */
+	@Deprecated
 	@Override
 	public KEY_GENERIC_CLASS[] toArray(KEY_GENERIC_CLASS[] input) {
-		if(input == null || input.length < length)
-		{
-			input = new KEY_GENERIC_CLASS[length];
+	final int length = size();
+		if(input == null || input.length < length) input = new KEY_GENERIC_CLASS[length];
+		if (start <= end) {
+			for(int i = 0,m=end-start;i<m;i++)
+				input[i] = KEY2OBJ(array[start + i]);
 		}
-		if (start <= end) System.arraycopy(array, start, input, 0, end - start);
 		else {
-			System.arraycopy(array, start, input, 0, length - start);
-			System.arraycopy(array, 0, input, length - start, end);
+			int offset = 0;
+			for(int i = 0,m=length-start;i<m;i++,offset++)
+				input[i] = KEY2OBJ(array[start + i]);
+			for(int i = 0;i<end;i++)
+				input[offset+i] = KEY2OBJ(array[i]);
 		}
 		return input;
 	}
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -18,7 +18,6 @@
 package PACKAGE;
 
 #if KEY_CLASS_Object
-import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Comparator;
 #endif
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -267,9 +267,9 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		return apparentLength >= 0 ? apparentLength : length + apparentLength;
 	}
 
+	/** Returns an array containing all of the elements in this PriorityQueue. */
 	@Override
-	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY()
-	{
+	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY() {
 		final int length = size();
 #if KEYS_PRIMITIVE
 		final KEY_GENERIC_TYPE[] newArray = new KEY_GENERIC_TYPE[length];
@@ -284,9 +284,9 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 		return newArray;
 	}
 
+	/** Returns an array containing all of the elements in this PriorityQueue. */
 	@Override
-	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input)
-	{
+	public KEY_GENERIC_TYPE[] TO_KEY_ARRAY(KEY_GENERIC_TYPE[] input) {
 		final int length = size();
 		if(input == null || input.length < length)
 		{
@@ -305,9 +305,9 @@ public class ARRAY_FIFO_QUEUE KEY_GENERIC implements PRIORITY_QUEUE KEY_GENERIC,
 	}
 
 #if KEYS_PRIMITIVE
+	/** Returns an array containing all of the elements in this PriorityQueue. */
 	@Override
-	public KEY_GENERIC_CLASS[] toArray(KEY_GENERIC_CLASS[] input)
-	{
+	public KEY_GENERIC_CLASS[] toArray(KEY_GENERIC_CLASS[] input) {
 		if(input == null || input.length < length)
 		{
 			input = new KEY_GENERIC_CLASS[length];
--- a/drv/ArrayFIFOQueue.drv
+++ b/drv/ArrayFIFOQueue.drv
@@ -18,6 +18,7 @@
 package PACKAGE;
 
 #if KEY_CLASS_Object
+import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Comparator;
 #endif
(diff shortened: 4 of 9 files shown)


Speiger:
@vigna @incaseoftrouble what about this patch?
This patch is honestly more important for this library then the list patch, for the following reasons:
- Priority Queues become usable in Serialization context. (And yes javas internal serialization context does not count)
- FIFOQueue becomes a option for LinkedLists use cases.
- HeapQueue becomes a usable option for TreeSets and sorted sets in general (with limitations but at least it becomes an option)

It fills a giant gap in Fastutils feature set.

drv/PriorityQueue.drv
```diff
@@ -60,6 +60,23 @@ public interface PRIORITY_QUEUE extends PriorityQueue<KEY_CLASS> {
...
+ 	 * <p>This default implementation just throws an {@link UnsupportedOperationException}.
+ 	 * @see #peek()
+	 * @param index of the element.
+	 * @return the indexed element.
+	 * @throws NoSuchElementException if the index is not in the range
+	 */
+
+	default KEY_GENERIC_TYPE PEEK(int index) { throw new UnsupportedOperationException(); }
+
+	boolean REMOVE_KEY(KEY_GENERIC_TYPE x);
```
incaseoftrouble, line 74: Add documentation here. This removes the *first occurence* of `x`, not all.

src/it/unimi/dsi/fastutil/PriorityQueues.java
```diff
@@ -63,12 +62,27 @@ public void clear() {}
...
+		public boolean remove(Object value) { return false; }
+
 		@Override
 		public void changed() { throw new NoSuchElementException(); }
 
 		@Override
 		public Comparator<?> comparator() { return null; }
 
+		@Override
+		public void trim() { throw new NoSuchElementException(); }
```
incaseoftrouble, line 75: should not throw an exception

src/it/unimi/dsi/fastutil/PriorityQueues.java
```diff
@@ -63,12 +62,27 @@ public void clear() {}
...
+
+		@Override
+		public Object peek(int index) { throw new NoSuchElementException(); }
+
+		@Override
+		public Object[] toArray() { throw new NoSuchElementException(); }
```
incaseoftrouble, line 81: should return empty array

src/it/unimi/dsi/fastutil/PriorityQueues.java
```diff
@@ -63,12 +62,27 @@ public void clear() {}
...
+
+		@Override
+		public Object[] toArray(Object[] input) { throw new NoSuchElementException(); }
```
incaseoftrouble, line 84: should return empty array

src/it/unimi/dsi/fastutil/PriorityQueue.java
```diff
@@ -76,6 +76,9 @@ default boolean isEmpty() {
 
 	void clear();
 
+	/** Trims the underlying array so that it has exactly {@link #size()} elements. */
+	void trim();
```
incaseoftrouble, line 80: Still doesn't belong here IMO

vigna: Absolutely, this has to go.

drv/PriorityQueue.drv
Speiger: Ok, wasn't sure what doc to add there. Thanks for that help

src/it/unimi/dsi/fastutil/PriorityQueue.java
Speiger: @incaseoftrouble @vigna ok i am kinda annoyed since that is a bit of a problem to me now.
What exactly are the rules that define that a function can not or can be added to a interface?
Because "All Implementations" using the function seems to be not the case.
Heck by the randomness you guys have provided so far i need to remove the following functions out of Priority Queue:
"Last, changed, comparator" and a few others.

I can understand it partly for lists since it is only implemented by the ArrayList, even so the size(int newSize) function exists.

Just my reasoning why trim function should exist:
- It is implemented in all underlying implementations that are not wrappers.
- It is the only way to "fast" shrink the list without using a while(!queue.isEmpty()) queue.dequeue(); Because clear function does not resize the array.
- It is really useful to have
- It is not accessible outside of casting which is not always possible since synchronizedwrappers exist.

It would be nice if a clear ruleset could be provided because this is getting to my nerves.
Also, "because it is not in java" is not really a good argument since "primitive lists" don't exist in the first place in java.

src/it/unimi/dsi/fastutil/PriorityQueues.java
Speiger: ok

src/it/unimi/dsi/fastutil/PriorityQueue.java
incaseoftrouble: Rules are very simple: Interfaces are independent of implementation. So only things which are tied to the concept are in the interface, nothing which is tied to a specific implementation thereof. This is not specific to fastutil or even Java, this is the whole point of interfaces, abstracting away from implementation. Example: The abstract concept of a (priority) queue is not bound to an array based implementation (or an array based implementation which is easy to trim). `trim()` does not make sense for this concept. Moreover, this is a public interface of a public library, you cannot possibly know all implementations. Take the Java Collection API as reference, a lot of smart people invested a lot of time in that. And yes it is a good argument since the underlying datastructures and algorithms are the same for lists, no matter if they are primitive or generic.

Why do you need to remove `last()` or `comparator()`?

`size(int)` is independent of the backing implementation. It has nothing to do with the backing array! In particular `size(0)` does not have to trim the array.

`clear()` could of course internally resize the array. Or add a check to remove operations if the current size is <= 1/4 of the array size. Heuristics like that are what keep the amortized runtime low. Calling `trim()` manually often is quite costly. There are very few situations where trim actually is useful. The only situation I can think of is that you heavily modify a list with a lot of removes and then keep a reference to it for a long time instead of copying it with a copy constructor.

Speiger: > Rules are very simple: Interfaces are independent of implementation. So only things which are tied to the concept are in the interface, nothing which is tied to a specific implementation thereof.

That is not a argument against adding a function like "trim". Which basically means: implementation please remove any excess you have.

> Why do you need to remove last() or comparator()?

Because they are specific to the "implementations" of the PriorityQueue. Last is only used by FIFO queue and comparator is only used by Heap/Array, and even sometimes not nessesary. (Nullable)

> size(int) is independent of the backing implementation. It has nothing to do with the backing array! In particular size(0) does not have to trim the array.

That is not true. size(int) is a function that could be compared with: "ensureCapacity(int requestedCapacity)" which is a ArrayList specific implementation. So if something like that comes though, why not trim? Because both do the same, just in opposite directions.

> clear() could of course internally resize the array

Yes it could. But it doesn't because a clear could also mean please clear the contents so i can refill it with something. A array change would basically cost a lot of performance. Which is the same reason why trim should exist. If a implementation for example does not shrink the array of a Queue (lets say a custom FIFO queue that only grows but not shrinks) could then only shrink with the "public" trim method. Which is only accessible if you make a non synchronized version or have a Reference object that you constantly need to manage too so that there is no multi threading issue.

Lets say it like this: Your arguments so far are very much collide with what is in the library already.
If a function like ensureCapacity (or size(int)) is already accepted but the opposite is rejected then remove the present one.

incaseoftrouble: > That is not a argument against adding a function like "trim". Which basically means: implementation please remove any excess you have.

That assumes implementations have any excess / don't remove them eagerly, which is not the case for, e.g., linked constructions. Also, as another argument, `trim()` does not have any observable effect (on the abstract interface) does not yield any result, its a no-op. Yes, memory is being cleared in some circumstances, but nothing changes in terms of the interface. Given the methods of the interface, there is no difference between the datastructure before and after trim. This screams "implementation dependence"

> Because they are specific to the "implementations" of the PriorityQueue. Last is only used by FIFO queue and comparator is only used by Heap/Array, and even sometimes not nessesary. (Nullable)

I don't get it. This MR does not remove any of these methods. But anyway, whether or not a Queue has a `last()` operation is up for debate. It usually rather belongs in Dequeue's. If a Queue has `last()` its also functioning as a stack (sort of, `pop` is missing). For `comparator()`, if you mean a getter method for the used comparator then it might well be in the interface but it doesn't have to. No reason to add but also no reason to remove. In any case, every implementation of priority queue needs some notion of priority (which can be canonically given as comparator).

> That is not true. size(int) is a function that could be compared with: "ensureCapacity(int requestedCapacity)" which is a ArrayList specific implementation. So if something like that comes though, why not trim? Because both do the same, just in opposite directions.

No. The documentation clearly states its effect in terms of the interface. In particular `size(int)` is *not* equivalent to `ensureCapacity`, `size(int)` drops exceeding objects. Actually you could even get a "soft" trim by calling `list.size(list.size())` (this is no argument for adding trim, this call does not force the implementation to do anything). I agree that it's a somewhat weird method to have though and I personally would remove it. However, it does satisfy the criteria of an interface method.

> But it doesn't because a clear could also mean please clear the contents so i can refill it with something. 

Yes, because in basically all applications the memory overhead is less important than the runtime overhead. Nothing stops you from subclassing the arraylist to eagerly trim the array though.

> An array change would basically cost a lot of performance.

`trim()` also needs to perform a linear copy, which is quite costly (even more costly than filling with nulls). Especially if you again add elements to the list.

> Which is the same reason why trim should exist. If a implementation for example does not shrink the array of a Queue (lets say a custom FIFO queue that only grows but not shrinks) could then only shrink with the "public" trim method. Which is only accessible if you make a non synchronized version or have a Reference object that you constantly need to manage too so that there is no multi threading issue.

What stops the implementation from being smart about trimming by itself? Just as it is smart about growing. That's the whole point of doubling the size of the array instead of growing it one at a time.

> Lets say it like this: Your arguments so far are very much collide with what is in the library already. If a function like ensureCapacity (or size(int)) is already accepted but the opposite is rejected then remove the present one.

As explained above, no. Also, I am not arguing that the choices of the library are all perfect, there are quite a few things that I'd remove / change, however modifying public API isn't something one should do without a lot of consideration.


Just as a question: What are the applications that you want to use `trim()` in and are you sure it helps you in any way? It hardly is relevant / often hurts more than it helps. Just for comparison, in the complete JDK, `trimToSize()` is only called two dozen times, while other array list modifications are called thousands of times.

Speiger: > That assumes implementations have any excess / don't remove them eagerly

That can be done with the clear function or any other function that could be interpreted different.
Not a argument.

> No. The documentation clearly states its effect in terms of the interface.

Honestly documentation and implementation is up for debate anyways. I mean the size function bug i am proposing a fix for.

> Yes, because in basically all applications the memory overhead is less important than the runtime overhead. Nothing stops you from subclassing the arraylist to eagerly trim the array though.

You have a point. But different when you think.

> What stops the implementation from being smart about trimming by itself? Just as it is smart about growing. That's the whole point of doubling the size of the array instead of growing it one at a time.

To clear the queue and then trim it to reduce arraycopy counts since resize is triggered every 1/4th of the array? If you wanted to shrink it down to the minimum size you would have to do 8 arraycopies during the dequeue loop. clear + trim would be 1 arraycopy. Which would be still smaller then the 8 array copies because you dont have to deal with the whiletrue loop just to trigger the 8 arraycopies. Ofc you could hack the system by doing: clear, enqueue single object and then dequeue. That could work, but then again thats a "Implementation specific hack". Which interfaces are there for so you don't need to do that.

> Just as a question: What are the applications that you want to use trim() in and are you sure it helps you in any way?

Work queue that you know can blow up sometimes and you want to keep it in check. Ofc a size parameter would be required in the function, but that is something that i use for example. I let it build up and every now and then again the trim function with a reasonable size gets triggered to ensure that the array if it exceeded the "normal" workrange would be cleared. 
Note that the Queue would be modified with the "protected" shrink function that does not trigger a shrink in the first place to optimize processing of tasks.

Anyways. This patch and all work i have done will be revoked by now because i am clear at that point that all work i will invest here will be included with a lot of waste of my time with the justification of why it would not possible to implement such features. I mean most of this took me a day or two of work to do. Just looking up implementations in the first place.
Its no wonder why these things are "still" missing.

I rather make my own branch and fix all the issue i found to have a library that does not make me scream "why the f... is this missing?". Which funnily enough is consistent with the entire repo.

Thanks you @incaseoftrouble for giving me a tutorial on how to actually work with this library. 
Sorry for being right now an asshole, but the trend shows me this is a waste of my time.

@vigna i highly suggest you start actually setting up rules and guide lines in a proper manner for features.
I know you don't have the time to code it yourself. But you are basically telling everyone (@incaseoftrouble helps with that FYI) who is willing to help and try to fix foundation issues that they should leave because they are not welcome.

Look at the pullrequests, how long people have to wait until they get something through. Even small things, is showing of that.
(if it was a large pullrequest i could understand it)

incaseoftrouble: > That can be done with the clear function or any other function that could be interpreted different. Not a argument.

No. The clear function specifies that, after calling, the collection is empty, i.e. `size()` yields 0, the iterator gives no elements etc. That is something entirely different than giving a hint to the implementation.

> To clear the queue and then trim it to reduce arraycopy counts since resize is triggered every 1/4th of the array? If you wanted to shrink a 1 Million Entries FIFO queue down to the minimum size you would have to do 8 arraycopies during the dequeue loop. clear + trim would be 1 arraycopy. Which would be still smaller then the 8 array copies because you dont have to deal with the whiletrue loop just to trigger the 8 arraycopies. Ofc you could hack the system by doing: clear, enqueue single object and then dequeue. That could work, but then again thats a "Implementation specific hack". Which interfaces are there for so you don't need to do that.

I don't get this. Why do I need 8 array copies? `clear()` can just eagerly empty the array. Then clearing even is a constant time operation (0 array copies!). Anyways, that was a suggestion, not a requirement.

> Work queue that you know can blow up sometimes and you want to keep it in check. Ofc a size parameter would be required in the function, but that is something that i use for example. I let it build up and every now and then again the trim function with a reasonable size gets triggered to ensure that the array if it exceeded the "normal" workrange would be cleared.

That's exactly what I am proposing the queue should do by itself :-) Once the queue is back to normal workrange (i.e. shrinking to a much smaller size than the array) it can just internally trim. 

> i highly suggest you start actually setting up rules and guide lines in a proper manner for features.

The things we are talking about really are not specific to this library. We discussed things like "what is an interface" and "what is whitespace".

Speiger: > Its only a "waste of time" because you are insisting on adding things which are unanimously rejected.

How long do you think the list patch would have waited for being imported? A month? 2? 6 weeks? 3 weeks?
Because I bet that that it would have been multiple weeks even after your review.

The problem I am having is very simple: It is hard to get started on the project because you have to have a certain set of knowledge to even start. Nothing to guide people who want to help.
When a feature request comes in, "I don't have the time but pullrequests are welcome",
And when they do it it stays around for months.
Without comment, but the "Moment" something comes in that is not welcome, it is very quickly pointed out.

Its fine that you don't want trim function in lists, i get that, hypocrisy ignored, priority queue i can get it too.
But with all  that combined. I get more things done in a month fixing it myself then having to wait a month on fastutil fixing foundation issues.
There is a lot of things that annoy me, but with all the things I would have done I would have waited a year to patch them all.

Speiger: I hope that helps you to understand @incaseoftrouble

incaseoftrouble: >  How long do you think the list patch would have waited for being imported? A month? 2? 6 weeks? 3 weeks?

Well, I got several thousand lines of code accepted in a few weeks. This patch could easily have been accepted this weekend.

>  Without comment, but the "Moment" something comes in that is not welcome, it is very quickly pointed out.

No, I just took time to look at this particular one because here I know where I can help.

> //Yeah totally not ensuring the capacity is right?

Did you even read the documentation?

> If the specified size is smaller than the current size, the last elements are discarded. Otherwise, they are filled with 0/{@code null}/{@code false}

This simply means that the implementation is wrong. Moreover, again, I never argued for the correctness of methods in fastutil. But even if some are disputable, this doesn't justify adding more things.

I still don't get your point about the queue. Why wouldn't `queue.clear()` trim the array? That's what I was suggesting. Implement `clear()` with
```java
public void clear() {
  this.a = ARRAYS.emptyArray(); // or however the reference to the empty arrays is called
  // size bookkeeping
}
```
voila. Or allocate the backing array to its default size.

Speiger: > Well, I got several thousand lines of code accepted in a few weeks. This patch could easily have been accepted this weekend.

I doubt it.

> No, I just took time to look at this particular one because here I know where I can help.

In this case vigna specifically. (Unless it has to be removed he barely answers or gives any indication of if something is good enough to be accepted. Only if something is wrong.)

> Did you even read the documentation?

Wait i am only writing the API not the implementation of things? If i am changing internals i am looking at the internals.
When i am proposing a new public API feature i am considering internal features and other API features (in combo how they are implemented) So in the grand view this seems more reasonable.

> I still don't get your point about the queue. Why wouldn't queue.clear() trim the array? That's what I was suggesting. Implement clear() with

The suggestion you do breaks any feature that uses a non reducing queue that wants to be used to be recycled so constant allocations are not often used. For example a Job System for path finding where the pathfinder recycles the queue so it does not allocate memory with every path finding attempt (or any other system that uses a recycled queue). This would break that.
Also every clear command would technically be a new object allocation. Even if it is a empty array.
And functions like clear are more carelessly used sine they are known to not allocate memory with every call. (Yes that could be accounted for)

My point being: You can not fix it with changing clear. You need a different method like "changed" to indicate you want to delete excess space. Maybe even the desired size. But that is rejected and no longer necessary since i have all that's needed.

incaseoftrouble: > I doubt it.

This is public information. Feel free to look at the MR's.

> The suggestion you do breaks any feature that uses a non reducing queue that wants to be used to be recycled so constant allocations are not often used.

Still has O(1) amortized runtime for addition and removal, compared to calling trim() from the outside without access to the size of the backing array, which could quickly degenerate into linear runtime for modifications. You need exponential growing / shrinking behaviour to keep array backed lists performant. And yes, this is why priority queue implementations do not offer this. Even C++ / Boost do not implement this behaviour or `trim()` for that matter. Because memory is typically isn't the bottleneck, especially at places where priority queues are used (work queues etc.).

> Wait i am only writing the API not the implementation of things? If i am changing internals i am looking at the internals.

The API (and its documentation) dictates the functionality. All implementations need to adhere to the API.

>  Also every clear command would technically be a new object allocation. Even if it is a empty array.

Nope, you don't need to re-allocate empty arrays, see above code example. But even if were a new object allocation, it would be, what, 16 bytes? Much faster than clearing the array in any case.

> My point being: You can not fix it with changing clear. 

My point is: There is nothing to fix. `trim()` is such a specialized operation which a) only applies to specific implementations and b) rarely is relevant and even then hardly does more good than harm. Its not a general purpose operation but a highly specific one, so just add a specialized implemenation for the specific use case (which can be done in a few minutes).

vigna: I made the same exact comment elsewhere for an analogous problem:

https://github.com/vigna/fastutil/pull/158#commitcomment-42962099

There's nothing bad in having trimming methods for specific implementations (e.g., hash tables for sets) but nobody would ever put trim() as a generic method for a dictionary.

Speiger: @vigna all implementations of PriorityQueue that are not wrapper implement trim, what exactly is bad in adding it to the interface? There is a reason such method exists.
I mean why provide an option if it is annoying to use a new feature added to fastutil.
And making a convince feature more accessible gets negative attention the moment they show up.
Why have them in the first place.

incaseoftrouble: We extensively explained this multiple times now. There are a multitude of reasons against it and I still don't see any argument for it except "I want it".

Speiger: Main reason is for giving the user of the library more control over how memory is managed, especially in the pooling range, without forcing them to implement custom shit for everything.

szarnekow: Chiming in here after following the discussion a little bit. I think the fact that the JavaDoc on the proposed `#trim()` is the only place in the interface that mentions the implementation detail `array` is already a good indicator that this would be misplaced there. PriorityQueues could potentially be implemented in different ways and imposing on implementors that they somehow must be able to `trim` to reduce memory footprint doesn't seem to be reasonable.
Users that want to have more control over the memory usage are always free to declare their types based on the implementation classes.

incaseoftrouble: > Main reason is for giving the user of the library more control over how memory is managed, especially in the pooling range, without forcing them to implement custom shit for everything.

Bloody hell. You *have* control already. You don't need to implement anything (except maybe a typed synchronization wrapper, which an IDE can generate with two clicks). ArrayList exposes trim already. If you are concerned about memory, you *need* to be implementation specific. Then use implementation specific types. Where is the problem?

Speiger: > Chiming in here after following the discussion a little bit. I think the fact that the JavaDoc on the proposed #trim() is the only place in the interface that mentions the implementation detail array is already a good indicator that this would be misplaced there. PriorityQueues could potentially be implemented in different ways and imposing on implementors that they somehow must be able to trim to reduce memory footprint doesn't seem to be reasonable.
Users that want to have more control over the memory usage are always free to declare their types based on the implementation classes.

So if a new implementation comes into play that does not have a array in the back. It could also just leave the function as noop. I mean, isn't that already done?

> Bloody hell. You have control already. You don't need to implement anything (except maybe a typed synchronization wrapper, which an IDE can generate with two clicks). ArrayList exposes trim already. If you are concerned about memory, you need to be implementation specific. Then use implementation specific types. Where is the problem?

Isn't the idea of this library that you don't have to custom implement needed but inaccessible features all the time?

Example: TreeSets not having poll functions, PriorityQueues missing remove/toArray/other functions, functions implemented in all childclasses but not accessible via interfaces, functions where you have to go left arm around the right ear to scratch the left ear?

incaseoftrouble: > So if a new implementation comes into play that does not have a array in the back. It could also just leave the function as noop. I mean, isn't that already done?

For the (it feels like) 10th time: This is not how interfaces work. If you want, an interface like
```java
ArrayBackedCollection {
  void trim();
  int arraySize();
}
```
might be possible. But still meh.

> Isn't the idea of this library that you don't have to custom implement needed but inaccessible features all the time?

They are all accessible :-)

> TreeSets not having poll functions

How would that even make sense? Its an ordered set, not a queue. You don't "poll", maybe `removeFirst()`.

> PriorityQueues missing remove/toArray/other functions

Because that's what defines (priority) queues. You put things in, you remove one from the front, that's it. That's what's described by the interface. Nothing more. If you want more, be more specific by moving to the specific type.

> functions implemented in all childclasses but not accessible via interfaces

You do not know all childclasses.

Speiger: > How would that even make sense? Its an ordered set, not a queue. You don't "poll", maybe removeFirst().

To be honest Pollfirst is a thing, but removeFirst too, but result is the same in that case.

> They are all accessible :-)

Yeah if you write your own wrappers or write handlers that lock existing wrappers to then access the nessesary function, because never saw that a synchronized queue could be casted.

> Because that's what defines (priority) queues. You put things in, you remove one from the front, that's it. That's what's described by the interface. Nothing more. If you want more, be more specific by moving to the specific type.

I don't know maybe you implement a memory manager that uses a Tombstsone system that fills its holes with last entries as long it is not a tombstone, but that would be only possible if you had a sortedSet or a PriorityQueue where you could delete entries but also properly poll. And since you love examples:
```java 
Int2IntMap slots = new Int2IntOpenHashMap();
IntArrayList slotData = new IntArrayList();
IntPriorityQueue freeSlots = new IntHeapPriorityQueue(); //Without my PriorityQueue fix this would be a Java TreeSet, because FastUtils implementations in that regard are either very laggy or not useable.
IntSet changes = new IntLinkedOpenHashSet();
	protected void clearEmptySlots()
	{
		if(freeSlots.isEmpty())
		{
			return;
		}
		if(slots.isEmpty())
		{
			freeSlots.clear();
			freeSlots.trim(); 
			slotData.clear();
			changes.clear();
			return;
		}
		int size = freeSlots.size();
		while(size > 0)
		{
			size -= ensureData();
			if(slotData.isEmpty() || freeSlots.isEmpty())
			{
				break;
			}
			int index = freeSlots.dequeueInt();
			int dataIndex = slotData.popInt();
			slotData.set(index, dataIndex);
			slots.put(dataIndex, index);
			changes.add(dataIndex);
		}
	}
	
	protected int ensureData()
	{
		int cleared = 0;
		while(slotData.topInt() == -1)
		{
			freeSlots.removeInt(slotData.size() - 1); //Referencing this here in that regard.
			slotData.popInt();
			cleared++;
		}
		return cleared;
	}
```

incaseoftrouble: > To be honest Pollfirst is a thing, but removeFirst too, but result is the same in that case.

What prevents you from doing `iterator().remove()` (like Java's tree set) or `remove(first())`? Feel free to add a default method to the interface and specialize it where applicable, shouldn't be that hard.

> Yeah if you write your own wrappers or write handlers that lock existing wrappers to then access the nessesary function, because never saw that a synchronized queue could be casted.

And what's the problem with generating a delegating synchronizing wrapper ("SynchronizedArrayList")? The few applications (dumb global collection synchronization (wtf!) + huge dataloads) need to optimize their implementation anyway / synchronize at other levels.

I have no idea what the API of your memory manager is supposed to be, I'd guess there are more fitting datastructures. But that's not the point here. Why are you not using fastutil's tree sets (they have removal, iterator based removal even fits the bulk removal style very well)? If they are slow, its not an issue of the API but their implementation. Also: Adding a remove to `PriorityQueue` is something worth considering, there isn't anything fundamentally wrong with that.

Speiger: > What prevents you from doing iterator().remove() (like Java's tree set) or remove(first())? Feel free to add a default method to the interface and specialize it where applicable, shouldn't be that hard.

TreeSet iterators from the center of the tree to the left to then remove it. And big changes are not unlikely to happen.
Big changes reference: A lot of entries getting deleted/inserted. Or a lot of small changes are happening at once and the array size would never exceed 10 element. Even so the flowthrough rate would be a lot bigger then that (1000-10000 times)

> And what's the problem with generating a delegating synchronizing wrapper ("SynchronizedArrayList")? The few applications (dumb global collection synchronization (wtf!) + huge dataloads) need to optimize their implementation anyway / synchronize at other levels.

PriorityQueue. You got that mixed up. Yeah Lets create a bunch of classes i expect the "Library" i use for that exact case to have.

> I have no idea what the API of your memory manager is supposed to be.

It is a memory aligned tombstone system that can grow and shrink but never leaves holes. And if holes are created they are shifted in from the end. 
Example: [0,1,2,3,4,5,6,7,8,9] => remove 4 and update => [0,1,2,3,9,5,6,7,8]. remove 5, and insert 16: [0,1,2,3,9,16,6,7,8]

The idea of that system is that i can have a Instanced GPU buffer to draw a lot of things much more faster and have offthread processing. I just need a system that keeps track of all the entries, and the class that takes care of that is the same class that this function came out of.

> Adding a remove to PriorityQueue is something worth considering, there isn't anything fundamentally wrong with that.

1 comment ago you said the opposite: > Because that's what defines (priority) queues. You put things in, you remove one from the front, that's it.

Speiger: Also lets provide an example of why I would use trim in the list interface.
this is an Annoyance since I have to hardcode everything on LongArrayLists instead of LongLists.
```java
	static final ThreadPool<LongArrayList> LISTS = new ThreadPool<LongArrayList>(100, LongArrayList::new, T -> {T.clear(); T.trim(1000);});
	World world;
	Long2ObjectMap<LongArrayList> ticks = new Long2ObjectOpenHashMap<LongArrayList>();
	Int2LongMap times = new Int2LongOpenHashMap();
	LongFilter filter = new LongFilter();
	
	public SingleTicketList(World world)
	{
		this.world = world;
	}
	
	@Override
	public synchronized boolean addTicket(int position, int data, int delay, ITicket ticket)
	{
		if(delay <= 0 || ticket.getId() < 0)
		{
			return false;
		}
		if(times.containsKey(position))
		{
			return false;
		}
		long time = world.getWorldTime() + delay;
		times.put(position, time);
		LongArrayList list = ticks.get(time);
		if(list == null)
		{
			list = LISTS.get();
			ticks.put(time, list);
		}
		list.add(BitUtil.toLong(position, (data << 4) | ticket.getId() & 15));
		return true;
	}

	@Override
	public synchronized void process(ITileMap map)
	{
		if(ticks.isEmpty())
		{
			return;
		}
		LongArrayList list = ticks.remove(world.getWorldTime());
		if(list == null)
		{
			return;
		}
		long[] array = list.elements();
		ITicket[] tickets = TicketRegistry.INSTANCE.getTickets();
		for(int i = 0,m=list.size();i<m;i++)
		{
			int data = BitUtil.toSecondInt(array[i]);
			tickets[data & 15].processTicket(map, world, BitUtil.toFirstInt(array[i]), data >> 4, this); 
		}
		LISTS.accept(list);
	}
```

incaseoftrouble: > When removing a object from a fastutil TreeSet it goes from center to left/right even so if it knows the position. And big changes are not unlikely to happen.

That's how `TreeSet#remove` works. TreeSets are awfully slow (`O(log n)` removal), that's nothing special about fastutil. You'd rather want a heap structure where you can bulk remove + heapify for large removals. If you use iterator removal its faster (for bulk operations). In general, you shouldn't use tree sets except you really really need the ordering because they are so slow.

> PriorityQueue. You got that mixed up. Yeah Lets create a bunch of classes i expect the "Library" i use to have in it.

No I didn't. General idea. I wouldn't expect a library to have such a thing because synchronizing on the collection level usually is a bad idea anyway if you are so dependent on the implementation (which means you got a special case going on).

> It is a memory aligned tombstone system that can grow and shrink but never leaves holes. And if holes are created they are shifted in from the end.

As I understand it, the holes are filled only on an "update" call (from the back) or when you add (filling from the front), right? Your implementation looks to be quite bloated if I understand it correctly. If this is performance critical, I would probably unwrap the array from the list to avoid costly vtable lookups and to have control over the memory usage. I'd probably use a (roaring)bitset to track free indices. Nothing more should be needed. Alternatively, a lazily sorted array could also work, depending on your read/write balance. I actually implemented something quite similar in JBDD, too (reference counting based). Works well with a few 10 million entries, but might not match your workload profile.

> 1 comment ago you said the opposite: "Because that's what defines (priority) queues. You put things in, you remove one from the front, that's it."

I didn't say that this speaks against adding the function, this is a mere explanation why it isn't there yet. `remove` simply is not a fundamental operation of priority queues. It is nice to have though, although it adds additional burden on implementations. Well ... you can probably `default` it with iterators.

>  this is an Annoyance since I have to hardcode everything on LongArrayLists instead of LongLists.

Well, this is exactly what you are supposed to do. You care about memory details. So you need to know the backing implementation.

Speiger: > That's how TreeSet#remove works. TreeSets are awfully slow (O(log n) removal), that's nothing special about fastutil. You'd rather want a heap structure where you can bulk remove + heapify for large removals. If you use iterator removal its faster (for bulk operations). In general, you shouldn't use tree sets except you really really need the ordering because they are so slow.

If i use a iterator in JavaTreeSet and remove any element it exactly knows where it is and only rebalances the tree and does not search for the value because it is known. Fastutil could have this optimization too since it keeps track of the current node. (Actually that was my second planned patch the moment both of the proposed patches came through)

> As I understand it, the holes are filled only on an "update" call (from the back) or when you add (filling from the front), right? Your implementation looks to be quite bloated if I understand it correctly. If this is performance critical, I would probably unwrap the array from the list to avoid costly vtable lookups and to have control over the memory usage. I'd probably use a (roaring)bitset to track free indices.

Didn't think of a bitset to use, could be great idea actually. It has auto sorting in natural numbers included and has 1/64th of the footprint for each index. Though I am not sure on the pollingperformance, compared to a priority queue. Especially if only a small amount of elements can be present at a given time. 

About bloating: The idea is that you have a memory mapper. I bet there are better implementations are out there, but what i achieved so far is pretty good in my eyes since in masschanging cases it performs really really well even if the cap of elements is 256 but 800 instances are active at the same time.

incaseoftrouble: > If i use a iterator in JavaTreeSet and remove any element it exactly knows where it is and only rebalances the tree and does not search for the value because it is known. Fastutil could have this optimization too since it keeps track of the current node. (Actually that was my second planned patch the moment both of the proposed patches came through)

Oh, the tree set iterator has non-optimized remove? Yeah that sucks. I didn't know that. That of course makes sense.

> Didn't think of a bitset to use, could be a great idea actually. It has auto sorting in natural numbers included and has 1/64th of the footprint for each index. Though I am not sure on the pollingperformance, compared to a priority queue. Especially if only a small amount of elements can be present at a given time. Since the usage could be either element 1-10 is empty or element 10000 => 250000 is empty. (not the amount but the range of where it is empty with 2-10 elements for example)

I can recommend `SparseBitSet` / `RoaringBitSet` (they are even faster and smaller). My `naturals-util` library has some wrappers around them to represent in a unified why. Feel free to try that one out. In particular, the bitset-backed sets extend fastutil's IntSet if I recall correctly, so they should be plug and play. Polling behaviour of bitsets is quite fast (especially for roaring, there its `O(1)`, java's BitSets however are pretty awful for sparse data). The cool thing is that bulk operations / range clears are extremely fast, too.

> Also the idea is that finding/removing a element is always the same speed (thats why the map, its dataValue => dataIndex and the list is the opposite) to prevent iteration. Ofc it has overhead but not much compared to forced iteration.

Ah, you remove by value, not by index, my mistake. Yes then you need mapping. But hash mapping is stupid fast anyway.

> Not really since it does not implement iterator.

Oh, true. Well then it indeed puts more load on the implementations. That would be an argument against it.

Speiger: > Oh, the tree set iterator has non-optimized remove? Yeah that sucks. I didn't know that. That of course makes sense.

Why do you think i am so frustrated with fastutil and want to get changes through so quickly, because I find all these issues with it. It is better for everyone if i do it myself for myself and then allow you guys to clone what you want. It makes live a lot easier for everyone.

> I can recommend SparseBitSet / RoaringBitSet (they are even faster and smaller). My naturals-util library has some wrappers around them to represent in a unified why. Feel free to try that one out. In particular, the bitset-backed sets extend fastutil's IntSet if I recall correctly, so they should be plug and play. Polling behaviour of bitsets is quite fast (especially for roaring, there its O(1), java's BitSets however are pretty awful for sparse data). The cool thing is that bulk operations / range clears are extremely fast, too.

Honestly PriorityQueue polling is fast enough in that regard. I mean it is shifting in small but a lot of changes at worst 2-4 elements around has no hashcheck and uses primitives. Speed in that regard is fast enough.
But I already looked into that library. I check out peoples code because curiosity about skill.

> Ah, you remove by value, not by index, my mistake. Yes then you need mapping. But hash mapping is stupid fast anyway.

Yep thats why it is Int2IntMap and IntList because thats the fastest way i found.

> Oh, true. Well then it indeed puts more load on the implementations. That would be an argument against it.

Yeah.... No Actually i will implement that into my version too, because that is a good idea.

incaseoftrouble: > Why do you think i am so frustrated with fastutil and want to get changes through so quickly, because I find all these issues with it. It is better for everyone if i do it myself for myself and then allow you guys to clone what you want. It makes live a lot easier for everyone.

I'd be happy about a proper implementation for that. This would very likely be accepted quite quickly if done right. API changes are much more delicate than implementation changes.

> Yeah.... No Actually i will implement that into my version too, because that is a good idea.

It is! But if you add it to the interface, it also has to be implemented in all the classes (= work), that's why simply adding it to the interface without default implementation might not be so good. It prevents people with custom implementations from updating if you don't provide a default implementation (arguably they could add an `UnsupportedOperationException`). It is a public library, its not just about your use-case :-)

Speiger: > I'd be happy about a proper implementation for that. This would very likely be accepted quite quickly if done right. API changes are much more delicate than implementation changes.

Again, i will do these changes for myself and allow the cloning of it. If you want to make a merge request yourself by cloning it and cut out unwanted changes you are welcome. But I am done with proposing things but then letting them decay over time. It is better for everyone and it keeps peace on both sides. Which is more desireable. I hope you agree
(the peek is one of the functions i tried to get into for a long enough time, to give credit and discredit at the same time. I was doing blind implementations for the longest time because i didn't know how to properly code with ,drv files)

>It is! But if you add it to the interface, it also has to be implemented in all the classes (= work), that's why simply adding it to the interface without default implementation might not be so good. It prevents people with custom implementations from updating if you don't provide a default implementation (arguably they could add an UnsupportedOperationException). It is a public library, its not just about your use-case :-)

How about let the interface implement "unsupported" instead of the implementation. Like by default don't expect this to work, (same could be said for trim xD, yes i am not leaving my stance here)

incaseoftrouble: > Again, i will do these changes for myself and allow the cloning of it.

That might not work out because of copyright / licences I guess.

> How about let the interface implement "unsupported" instead of the implementation.  Like by default don't expect this to work,

This definitely isn't good style. The point of interfaces is to describe the capabilities of an object. Adding optional features should rather be done by extending the interface (See one of the biggest fuck-ups of Java: Allowing to throw such exceptions in the Collections API for immutable collections instead of introducing `MutableXY`. Another example why you should really be careful with API changes. You can't really revert them.)

Speiger: > That might not work out because of copyright / licences I guess.

Technically it does, because there is written permission out there now. As long I don't revoke it, it is technically valid.
And I can not retroactively remove the permission. (In other words if i say no, it is anything after that)

PR #169, opened 2020-10-24
Improvements to Lists

Speiger:
A more improved List Patch.
All functions changes are limited to List, ArrayList, Lists classes and referencing test classes.

As suggested from incaseoftrouble to improve the merging process.

Table of Contents:
- Added: reverse function for Lists that also checks for random access.
- Fixed: Unmodifiable list was modifiable via the size function.
- Added: Improved implementation of removeIf to ArrayList
- Added: Improved implementation of retainAll to ArrayList.
- Added: Tests for all new functions

Pull request is considered ready for review from my side.

### DIFF
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -570,6 +570,20 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		size = j;
 		return modified;
 	}
+	
+	@Override
+	public boolean retainAll(final COLLECTION c) {
+		final KEY_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (c.contains(a[i])) a[j++] = a[i];
+#if KEYS_REFERENCE
+		Arrays.fill(a, j, size, null);
+#endif
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
 
 #endif
 
@@ -586,6 +600,59 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		size = j;
 		return modified;
 	}
+
+	@Override
+	public boolean retainAll(final Collection<?> c) {
+		final KEY_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (c.contains(KEY2OBJ(a[i]))) a[j++] = a[i];
+#if KEYS_REFERENCE
+		Arrays.fill(a, j, size, null);
+#endif
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+
+#if KEYS_REFERENCE
+	@Override
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {
+		java.util.Objects.requireNonNull(filter);
+		final KEY_GENERIC_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		Arrays.fill(a, j, size, null);
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+#elif KEYS_PRIMITIVE
+#ifdef JDK_PRIMITIVE_PREDICATE
+	/** Remove from this collection all elements which satisfy the given predicate.
+	 *
+	 * @param filter a predicate which returns {@code true} for elements to be
+	 *        removed.
+	 * @see Collection#removeIf(java.util.function.Predicate)
+	 * @return {@code true} if any elements were removed.
+	 */
+	@Override
+	public boolean removeIf(final JDK_PRIMITIVE_PREDICATE filter) {
+		java.util.Objects.requireNonNull(filter);
+		final KEY_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+	
+	@Override
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {return removeIf((JDK_PRIMITIVE_PREDICATE) key -> filter.test(KEY2OBJ(KEY_NARROWING(key))));}
+#endif
+#endif
 
 	@Override
 	public KEY_LIST_ITERATOR KEY_GENERIC listIterator(final int index) {
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -559,6 +559,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 
 	@Override
 	public boolean removeAll(final COLLECTION c) {
+		Objects.requireNonNull(c);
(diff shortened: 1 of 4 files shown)


drv/ArrayList.drv
```diff
@@ -571,6 +571,20 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
+	@Override
+	public boolean retainAll(final COLLECTION c) {
+		final KEY_TYPE[] a = this.a;
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (c.contains(a[i])) a[j++] = a[i];
+#if KEYS_REFERENCE
+		Arrays.fill(a, j, size, null);
+#endif
+		final boolean modified = size != j;
```
incaseoftrouble, line 583: @vigna More general question: Maybe the list should trim() here if it significantly decreased in size?

drv/ArrayList.drv
```diff
@@ -587,6 +601,59 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		return modified;
 	}
 
+	@Override
+	public boolean retainAll(final Collection<?> c) {
```
incaseoftrouble, line 605: If `c` is an instance of `COLLECTION` this should delegate to the type-specific call to avoid boxing

drv/ArrayList.drv
```diff
@@ -571,6 +571,20 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		return modified;
 	}
 
+	@Override
+	public boolean retainAll(final COLLECTION c) {
+		final KEY_TYPE[] a = this.a;
```
incaseoftrouble, line 576: for consistency with the `removeIf` etc., I guess `java.util.Objects.requireNonNull(c);`

drv/ArrayList.drv
```diff
@@ -587,6 +601,59 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
+		int j = 0;
+		for(int i = 0; i < size; i++)
+			if (!filter.test(a[i])) a[j++] = a[i];
+		final boolean modified = size != j;
+		size = j;
+		return modified;
+	}
+
+	@Override
+	public boolean removeIf(final java.util.function.Predicate<? super KEY_GENERIC_CLASS> filter) {return removeIf((JDK_PRIMITIVE_PREDICATE) key -> filter.test(KEY2OBJ(KEY_NARROWING(key))));}
```
incaseoftrouble, line 653: return probably should be on a new line, right?

drv/Lists.drv
```diff
@@ -47,6 +47,26 @@ public final class LISTS {
...
+			for (int i=0, mid=size>>1, j=size-1; i<mid; i++, j--) {
+				final KEY_GENERIC_TYPE t = l.GET_KEY(i);
+				l.set(i, l.GET_KEY(j));
+				l.set(j, t);
+			}
+			return l;
+		}
+		KEY_LIST_ITERATOR KEY_GENERIC fwd = l.listIterator();
+		KEY_LIST_ITERATOR KEY_GENERIC rev = l.listIterator(size);
+		for (int i=0, mid=size>>1, j=size-1; i<mid; i++, j--) {
```
incaseoftrouble, line 62: I think this should be `size >>> 1` (not `size >> 1`). However this won't have any practical impact.

Also: Formatting. Should have some more spaces :-)

test/it/unimi/dsi/fastutil/ints/IntArrayListTest.java
```diff
@@ -61,14 +61,25 @@ public void testRemoveAll() {
 		assertEquals(IntArrayList.wrap(new int[] { 0, 2 }), l);
 	}
 
+	@Test
+	public void testRetainAll()
+	{
```
incaseoftrouble, line 66: brace on previous line

test/it/unimi/dsi/fastutil/ints/IntListsTest.java
```diff
@@ -45,4 +46,12 @@ public int size() {
 		assertTrue(IntLists.synchronize(new IntArrayList()) instanceof RandomAccess);
 		assertTrue(IntLists.synchronize(new IntArrayList(), new Object()) instanceof RandomAccess);
 	}
+
+	@Test
+	public void testReverse()
+	{
```
incaseoftrouble, line 52: brace on previous line

incaseoftrouble:
looks good overall, only minor things IMO 👍

drv/ArrayList.drv
Speiger: Should I also add this to removeAll? because this was just a removeAll copy without the "!" at the contains check.

drv/ArrayList.drv
Speiger: Technically you would have to define it as collection<?> first to actually cause a case like that and have the boxing problem beforehand.
On top of that same question again: should this apply to removeAll too? Because it is just a copypaste job and removing 1 character from the function.

drv/ArrayList.drv
Speiger: Not in the cases I have seen. (any synchronization wrapper class does the exact same thing)

drv/Lists.drv
Speiger: To be fair this is about negative values. If you have those you have other problems.
Thanks.

test/it/unimi/dsi/fastutil/ints/IntArrayListTest.java
Speiger: Oh I missed that one. Thanks

Speiger:
Thanks.

drv/ArrayList.drv
incaseoftrouble: > Technically you would have to define it as collection<?> first to actually cause a case like that and have the boxing problem beforehand.

What? As soon as this method is entered there is boxing.

Yeah I guess it makes sense to have it in removeAll, too.

drv/ArrayList.drv
incaseoftrouble: Ah I see.

drv/Lists.drv
incaseoftrouble: Oh wait nevermind, its not relevant here, only for binary search.

drv/ArrayList.drv
incaseoftrouble: Yeah I think so.

drv/ArrayList.drv
Speiger: You have to define it a List<Integer> instead of a IntList to enforce  that Collection<Integer> is called instead of IntCollection.
How do you test in fastutil for a collection type?

incaseoftrouble: Well, yes, but if `IntList` is used as an implementation of `Collection<Integer>` for example, this method is called.

I guess `if (c instanceof COLLECTION)`?

Speiger: If it is defined as Collection yes, otherwise the most top interface applies and since IntCollection is above Collection that one gets used. You have to choose to do it is all i am saying.

incaseoftrouble: Huh? Above? The reference is resolved at compile time, the most specific method is used. IntCollection is "below" Collection in the sense that it is more specific.

You don't necessarily have to choose to do so when you as soon as you are working on either side of an API where the other side doesn't know about fastutil.

Anyway, the check is cheap for a bulk operation, so it won't hurt to have it there I guess.

PR #176, opened 2020-11-03
Main method test integration

techsy730:
Include the optionally generated tests in the `main` methods (if present) in the regular unit tests (generated by passing `TEST=1` to the `make` command). If `TEST=1` is not included and thus these `main` methods are not generated, then these new test methods gracefully fallback to doing nothing and nothing changes from the current status quo.

These legacy main method tests contain a large battery of tests. This allows the main unit tests to pick them up without having to go through the monumental effort of migrating all of them to normal unit tests. This allows, for example, cleaner gathering of code coverage numbers.

This approximately doubles the code coverage when running unit tests.

To accomplish this, the main method tests were updated to throw exceptions instead of a clean return with a `println`, or worse, `System.exit(1)`.
Also fixes some issues with the tests, such as some cases of quadratic scaling and tests on the wrong things.

### Before

ALL: 2.2%

ints package: 10.9%
Int2IntAVLTreeMap: 29.5%
Int2IntOpenHashMap: 45.4%
Int2IntRBTreeMap: 31.2%
IntArrayList: 16.4%
IntAVLTreeSet: 0.0%
IntOpenHashSet: 32.4%
IntRBTreeSet: 17.3%

objects package: 2.0%
Object2IntOpenHashMap: 22.8%
ObjectArrayList: 15.7%
ObjectAVLTreeSet: 1.9%
ObjectOpenHashSet: 27.1%
ObjectRBTreeSet: 17.5%

### After

ALL: 3.9%

ints package: 19.8%
Int2IntAVLTreeMap: 75.3%
Int2IntOpenHashMap: 72.2%
Int2IntRBTreeMap: 77.0%
IntArrayList: 88.5%
IntAVLTreeSet: 75.9%
IntOpenHashSet: 67.3%
IntRBTreeSet: 79.9%

objects package: 4.8%
Object2IntOpenHashMap: 43.3%
ObjectArrayList: 86.4%
ObjectAVLTreeSet: 76.2%
ObjectOpenHashSet: 57.7%
ObjectRBTreeSet: 80.2%


incaseoftrouble:
Is there any reason to keep the `main` methods

techsy730:
Yes, these tests should be migrated to proper unit tests over time (and the benchmarks probably pulled into a separate file). But that would take considerably more time then I put into this PR.
I figured this would be a good way to fill this hole until that is done. I should make that a bit more clear in the description and TODO.

vigna:
The main tests come from some from obscure, dark days (end of 90's, anyone?). It was a different world. Yes, in time everything should be migrated to proper unit tests, but let's focus on the features now (and NavigableSet/NavigableMap are still on the TODO list, sigh).

vigna:
Argh, I should have merged this before the last round of changes. I eliminated valEquals() if favor of Objects.equals() everywhere and now it's a gigantic conflict. 🤦🏻‍♂️

techsy730:
Ah, no problem. It looked worse then it really was.

PR #178, opened 2020-11-09
Update object based bulk operation methods to use type specific

techsy730:
Update object based, boxing bulk operation methods to use the fast type-specific version if possible.

AbstractCollection.addAll
                  .containsAll
                  .removeAll
                  .retainAll
AbstractList.addAll(int, Collection)
AbstractSet.equals

Will now delegate to the fast, type-specific method if given an type-specific collection as their input.

This is a slight behavior change. If there are any user subclasses of AbstractCollection/AbstractList that have implemented their type specific bulk methods as calling the object based ones (despite that being strongly dis-recommended) without overriding the boxed versions being overridden too, this will now cause an infinite recursion.

### DIFF
--- a/drv/AbstractCollection.drv
+++ b/drv/AbstractCollection.drv
@@ -18,6 +18,7 @@
 package PACKAGE;
 
 import java.util.AbstractCollection;
+import java.util.Collection;
 
 /** An abstract class providing basic methods for collections implementing a type-specific interface.
  *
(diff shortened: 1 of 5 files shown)


vigna:
It's a good idea—in fact, I'm puzzled why it isn't already like that. Too many methods to think of and optimize, I guess.

PR #180, opened 2020-12-14
Harden immutability of ImmutableList.

techsy730:
Harden immutability of ImmutableList.

This closes a few avenues that could have lead to mutating the supposed to be immutable list. In particular, the backing array was non-final and protected (now private and final), and mutations could still happen through the iterator.

Adds the cult-favorite optimization of sharing a single empty instance if given empty arguments to the static constructor.

Also add fast toArray for object based ImmutableList and ArrayList.

### DIFF
--- a/drv/ImmutableList.drv
+++ b/drv/ImmutableList.drv
@@ -279,7 +279,9 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	@Override
 	public KEY_GENERIC KEY_GENERIC_TYPE[] toArray(KEY_GENERIC_TYPE a[]) {
-		if (a == null || a.length < size()) {
+		if (a == null) {
+			a = KEY_GENERIC_ARRAY_CAST new Object[size()];
+		} else if (a.length < size()) {
 			a = KEY_GENERIC_ARRAY_CAST Array.newInstance(a.getClass().getComponentType(), size());
 		}
 		System.arraycopy(this.a, 0, a, 0, size());
--- a/drv/ImmutableList.drv
+++ b/drv/ImmutableList.drv
@@ -293,16 +293,16 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
 		ensureIndex(index);
 
 		return new KEY_LIST_ITERATOR KEY_GENERIC() {
-				int pos = index, last = -1;
+				int pos = index;
 
 				@Override
 				public boolean hasNext() { return pos < a.length; }
 				@Override
 				public boolean hasPrevious() { return pos > 0; }
 				@Override
-				public KEY_GENERIC_TYPE NEXT_KEY() { if (! hasNext()) throw new NoSuchElementException(); return a[last = pos++]; }
+				public KEY_GENERIC_TYPE NEXT_KEY() { if (! hasNext()) throw new NoSuchElementException(); return a[pos++]; }
 				@Override
-				public KEY_GENERIC_TYPE PREV_KEY() { if (! hasPrevious()) throw new NoSuchElementException(); return a[last = --pos]; }
+				public KEY_GENERIC_TYPE PREV_KEY() { if (! hasPrevious()) throw new NoSuchElementException(); return a[--pos]; }
 				@Override
 				public int nextIndex() { return pos; }
 				@Override
@@ -417,6 +417,11 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
 			this.pos = myNewPos;
 			return new Spliterator(oldPos, retMax);
 		}
+	}
+
+	@Override
+	public KEY_SPLITERATOR KEY_GENERIC spliterator() {
+		return new Spliterator();
 	}
 
 	@Override
--- a/drv/ImmutableList.drv
+++ b/drv/ImmutableList.drv
@@ -48,6 +48,19 @@ import java.util.stream.Collector;
 
 public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements RandomAccess, Cloneable, java.io.Serializable {
 	private static final long serialVersionUID = 0L;
+	
+	SUPPRESS_WARNINGS_KEY_UNCHECKED_RAWTYPES
+	private static final IMMUTABLE_LIST EMPTY = new IMMUTABLE_LIST(ARRAYS.EMPTY_ARRAY);
+
+#if KEYS_PRIMITIVE
+#define _EMPTY_ARRAY ARRAYS.EMPTY_ARRAY
+#else
+	SUPPRESS_WARNINGS_KEY_UNCHECKED
+	private static final KEY_GENERIC KEY_GENERIC_TYPE[] emptyArray() {
+		return KEY_GENERIC_ARRAY_CAST ARRAYS.EMPTY_ARRAY;
+	}
+#define _EMPTY_ARRAY emptyArray()
+#endif
 
 	/** The backing array; all elements are part of this list. */
 	private final KEY_GENERIC_TYPE a[];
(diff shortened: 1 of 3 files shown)


vigna:
I'm a bit confused. What does this defend us against? The user can always subclass and modify the underlying array—it's protected. Waiting for sealed classes (http://openjdk.java.net/jeps/360) shouldn't we make the backing array private? And at that point, wouldn't all mutation method inducing an actual mutation throw an exception anyway? It's true that in this case if one does an addAll(emptyset) there is no thrown exception, but this seems to me the only case covered here. Am I wrong?

techsy730:
Oh, I didn't realize the backing array was protected, not private.

If you think overriding all of those methods to throw explicitly is too much noise, I'll remove them. (Also, addAll(empty collection) is fine to not throw as now mutation is attempted)

vigna:
Actually,  it could be a good idea to make it private. No? I think the JDK throws exceptions on any mutation method,  _even if the method would not cause a mutation_.

techsy730:
The reason I overrode the methods is so if you did something like.

ImmutableIntList intList = ImmutableIntList.of(2,3,4);
intList.add(5);
The add method would show up as deprecated and strikethrough as the compiletime type of intList if ImmutableIntList. If you don't think that is worth their weight I will remove them.

incaseoftrouble:
> I think the JDK throws exceptions on any mutation method, even if the method would not cause a mutation.

Yes and its a very good idea to have that IMO. If something is immutable, it should be verbose about it.

vigna:
Just thinking. This might  be a good idea but it should happen for all immutable classes for ortogonality. Probably only ImmutablePair needs fixing.

techsy730:
I'm working on the bulk operation methods now and am also adding the cult-favorite optimization of sharing an empty instance from the static factory.

techsy730:
@vigna If we are going that route, then we should an ImmutableCollection class be made that both of them extend (which will also shoulder some of the burden of some of the overrides)

vigna:
I'm thinking. An AbstractImmutableCollection extending AbstractCollection would contain basically the same methods and work for future immutable sets, for example. We cannot possibly have completeness—if the JDK adds default mutation methods to the Collection interface we will not cover that, but it does not seem so bad. 

I also realized that Pair has two types and it does not make any sense to make into a collection. SortedPair might be a candidate, but the main usefulness I see is contains(), which would tell you whether either element in the pair is equal to some other element. But in order to implement that, we should implement a lot of boilerplate code (iterators, etc.) and it seems a bit overkill to me. Or we should have an additional SameTypePair interface which extends Pair<K,K> and implement sthose as default methods.

I just don't know if this is worth the additional complexity.

I'm adding now an ImmutableCollection driver file—would you kindly move the deprecated/finalized mutation methods up there? Thanks!

incaseoftrouble:
> I also realized that Pair has two types and it does not make any sense to make into a collection.

Jeah, also conceptually it doesn't make sense IMO. A collection is an arbitrary length structure. A pair is a 2-tuple / 2-list, not a list.

vigna:
Meh. In standard mathematics, a pair is a set. In the standard axiomatization of mathematics (Zermelo–Fraenkel or Von Neumann), you define an unordered pair as {x, y}, and an ordered pair as { x, y, {x} }. An unordered pair is a set containing the elements, and an ordered pair is a set containing the elements of the pair and a singleton set that tells you which is the first element of the pair. So it is not completely nonsensical. But I agree it's a stretch. We might add just a default convenience method contains() to SortedPair.

vigna:
@techsy730 no I completely forgot about the lack of multiple inheritance. Sets inherit from AbstractSet, lists from AbstractList, etc. There's no way to slip an class with finalized, deprecated mutation methods in. The best we can do is an AbstractImmutableList extending AbstractLlist, but that really feels wrong.

incaseoftrouble:
> So it is not completely nonsensical.

Not mathematically, but for JDK's notion of a "Collection".

`contains` certainly makes sense! But not the whole barrage of collection methods I think.

incaseoftrouble:
> no I completely forgot about the lack of multiple inheritance. Sets inherit from AbstractSet, lists from AbstractList, etc. There's no way to slip an class with finalized, deprecated mutation methods in.

You could add mixin-interfaces with default methods maybe? But that's also meh

> The best we can do is an AbstractImmutableList extending AbstractLlist, but that really feels wrong.

Why? Otherwise you have ConcreteImmutableList extends AbstractList, right :-)?

vigna:
> You could add mixin-interfaces with default methods maybe? But that's also meh

Wait wait. Can you really do that? Override already-implemented methods using a default method from an interface?

> > The best we can do is an AbstractImmutableList extending AbstractLlist, but that really feels wrong.
> 
> Why? Otherwise you have ConcreteImmutableList extends AbstractList, right :-)?

Wrong in the sense that AbstractList > AbstractImmutableList > ImmutableList and then there is no other immutable list implementation seems overkill...

incaseoftrouble:
> Wait wait. Can you really do that? Override already-implemented methods using a default method from an interface?

Oh, not if they are implemented, I think, true.

> Wrong in the sense that AbstractList > AbstractImmutableList > ImmutableList and then there is no other immutable list implementation seems overkill...

Ah, true. Yeah.

techsy730:
One possibility is to not have ImmutableList extend AbstractList, and just implement everything as needed (sublist wouldn't be too terribly hard, as it is just a view over a subarray, and the rest would be busywork method forwarding). That way it can extend ImmutableCollection.

But if we have only one concrete Immutable type, is that refactor really worth it?

techsy730:
> Wait wait. Can you really do that? Override already-implemented methods using a default method from an interface?

Sort of. You have to do it explicitly, but you can do

```java
class A extends BFoo implements IFoo {
  @Override
  public void foo() {
    IFoo.super.foo();
  }
}```
and I believe that makes foo call IFoo's default implementation instead of BFoo's implementation (which would normally win otherwise)

**Comment by incaseoftrouble** [comment]
> and I believe that makes foo call IFoo's default implementation instead of BFoo's implementation (which would normally win otherwise)

yes, that works. But then you might aswell write
```java
  @Override
  public void foo() {
    throw immutable("foo");
  }
```
(where `immutable` is some helper generating the exception)

techsy730:
Well, yes, for our case we would be better just throwing the exception or making a separate helper method. But I was speaking in general where such a thing might not be so easy.

vigna:
> > and I believe that makes foo call IFoo's default implementation instead of BFoo's implementation (which would normally win otherwise)
> 
> yes, that works. But then you might aswell write
> 
> ```java
>   @Override
>   public void foo() {
>     throw immutable("foo");
>   }
> ```
> 
> (where `immutable` is some helper generating the exception)

But this is even _more code_ 😂

techsy730:
Re: Pair as collection, while implementing Collection is probably indeed overkill, it might be worth implementing Iterable?

drv/ImmutableList.drv
```diff
@@ -436,5 +481,378 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
...
+	 *
+	 * @deprecated
+	 */
+	@Override
+	@Deprecated
+	public final void add(final int index, final KEY_GENERIC_TYPE k) {
+		throw new UnsupportedOperationException();
+	}
+
+	/**
```
vigna, line 497: I think these should be all {@inheritDoc} at the start and have @implSpec tagging the implementation specification.

incaseoftrouble:
> Re: Pair as collection, while implementing Collection is probably indeed overkill, it might be worth implementing Iterable?

Well, what would be its type? Pair has (in general) two distinct elements and `Iterable<Object>` isn't that useful. Maybe you can add something like `public static <TYPE, LEFT extends TYPE, RIGHT extends TYPE> List<TYPE> asList(Pair<LEFT, RIGHT>)`?

vigna:
Guys, can we just forget this stupid idea of mine? 😂  I'll add a contains() method to SortedPair.

vigna:
@techsy730 do you have deprecation warnings on? I'm finding a lot of BIG_ARRAYS.get(), etc. instead of BigArrays.get() around.

vigna:
Also, at line 473 of BigListIterators the max variable is never used—can you check it can be removed?

vigna:
I'm cumulating here observations about the whole code checked in the the last few days, as this is is open (I know, not good practice, please bear with me).

I'm seeing this:
```
   [javac] /Users/vigna/git/fastutil/src/it/unimi/dsi/fastutil/floats/FloatIterable.java:130: warning: [overloads] forEach(DoubleConsumer) in FloatIterable is potentially ambiguous with forEach(Consumer<? super Float>) in FloatIterable
    [javac] 	default void forEach(final java.util.function.DoubleConsumer action) {
    [javac] 	             ^
```

That does not sound good. What is happening?

vigna:
OK, 409 warnings still. This stuff needs some work.

techsy730:
Mind if I deal with the compile warnings in another PR?

vigna:
Not at all. I fixed a lot of those,  but there are a few unused private method and classe and I don't want to mess up your work.

PR #184, opened 2020-12-17
Fix an unused class and some unused imports.

techsy730:
A few more warnings.


vigna:
We stepped a bit on each other toes here. In general, when the logical condition for an import is crazy and the usage is low, I prefer to simply add the full package specification in all occurrences.

Issue #183, opened 2020-12-17
[OPTIMIZATION] Methods that return 3 or more possible runtime-types optimize very poorly


techsy730:
Something discovered when optimizing Guava is that Hotspot (OpenJDK's virtual machine) optimizes megamorphic variables/returns (those that could be 3 or more different runtime types) _far_ worse then it does mono or dimorphic ones (1 or 2 different implementations). Sometimes by factors in the double digits. (https://github.com/google/guava/issues/1268)

As such, we should avoid static utility methods that can return potentially 3 or more different implementations.

This may (or may not) be better in newer JVMs, but in Java 8, the one we are targeting, it is absolutely an issue.

vigna:
Mmmmhhh... I think there are a lot of those. Like, the various .of() methods which optimize for an empty set, a singleton, etc.

techsy730:
There are two offenders that immediately come to mind
* The `of` static factories from the root interfaces (`List.of(...)`, etc)
* The wrappers of Spliterator in `Spliterators` converting to/from object based and primitive iterators

techsy730:
The `of` methods can be resolved by having the empty case return a "well known" empty instance of the kind the varargs one returns. Such as `ImmutableList.EMPTY` instead of `Lists.EMPTY` for `List.of()`.

The `Iterators` case will probably require me to flatten the class hierarchy, pulling the `instanceof` into `IteratorWrapper` instead of having a subclass of it. (would want to measure this impact)

vigna:
Yeah, I thought about that. We can have an immutable empty list exposed directly by ImmutableList. I'm writing that.

techsy730:
I've already done that in https://github.com/vigna/fastutil/pull/180

vigna:
Great! :) I'm still finding unused classes like

PrimitiveSpliteratorWrapperWithComparator

in ByteSpliterators.

techsy730:
Oops, I missed those. Forgot to check the shorter types' warnings. I'll fix those

techsy730:
List.of and Set.of are the most important.

Iterators.asXIterator and its cousin Spliterators.asXSpliterator 
However, the performance loss only happens if three or more returns actually happen in a program run, not just possible at all. And I can't imagine that many programs will trigger all three paths. Still should be looked at but less important.

techsy730:
I should try to measure which has the worst impact performance wise:
1. The current status quo (megamorphic return)
2. Moving the `instanceof` check into the nextInt() method
[...]

PR #185, opened 2020-12-17
Make List.of and Set.of bimorphic

techsy730:
Make List.of and Set.of bimorphic, that is, only returns potentially two different types. Three or more tanks performance severely.
See https://github.com/vigna/fastutil/issues/183

Now
* List.of will return one of ImmutableList or Lists.Singleton
* Set.of will return one of Sets.UnmodifiableSet or Sets.Singleton, and in the former case the inner set will be one of ArraySet or OpenHashSet.

There is one more case that needs to be looked into, and that is Iterators.asXIterator and its cousin Spliterators.asXSpliterator.
However, the performance loss only happens if three or more return types actually happen in a program run, not by merely being possible in the bytecode. And I can't imagine that many programs will trigger all three paths. Still should be looked at but less important.


techsy730:
With these final few pull requests (especially this one, the bimorphic call for `of`), I think I am done for this release. Obviously I can do further work for later releases, but IMHO these gets us in a good spot for a release.

vigna:
I agree. My only wish was for NavigableSet, we're really late on that, but it's a lot of work.

techsy730:
I can try doing that for 8.6. Since I am going to have to break down and understand the tree structure anyways to make an efficient spliterator, it would also get me the knowledge needed to implement the navigable methods.

vigna:
We can work on it together for 8.6.0, it would be a nice addition (and the last before switching to Java 9 or even 11). For me, the main benefit is upper/lower bound computations—all the reversed view stuff I think is really a red herring (and a lot of work).

PR #186, opened 2020-12-18
Make forEachTYPE instead of overloading forEach for primitive collections

techsy730:
Rename the forEach(TYPEConsumer) methods to forEachTYPE(TYPEConsumer) for primitive TYPEs.

This solves an ambiguity when calling forEach with a lambda requiring casting the lambda to solve.
By happenstance it wasn't an issue for int, long, and double collections due to Fastutil's consumer being both an object and primitive consumer, solving the ambiguity. However, it was an issue for byte, short, char, and float collections.

This also means we can remove the messy "don't override this" disambiguation overrides that int, long, and double Iterables had, and cleans up all the cruft associated with it.

Note this only fixes it for type specific `Iterable`s, not for the `Spliterator`s. However, as Spliterators are rarely used "raw", this isn't as big of an issue. Perhaps can be a future cleanup.

### DIFF
--- a/drv/AbstractBigList.drv
+++ b/drv/AbstractBigList.drv
@@ -247,13 +247,13 @@ public abstract class ABSTRACT_BIG_LIST KEY_GENERIC extends ABSTRACT_COLLECTION 
 	 * to using the iterator based loop implementation from the superinterface.
 	 */
 	 @Override
-	 public void forEach(final METHOD_ARG_KEY_CONSUMER action) {
+	 public void FOREACH_KEY(final METHOD_ARG_KEY_CONSUMER action) {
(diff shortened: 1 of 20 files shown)


vigna:
@techsy730 I'm in awe by the speed and precision with which you code, but can we _talk_ before you write hundreds of line of code? 😂  You totally convinced me there was no need for this. LOL.

vigna:
So, I checked every possible case I could think of: I must conclude that the warning from javac is wrong. No matter what I did—lambdas, variables—I could not create an ambiguous call (in Eclipse—in the past, I had the two compilers diverge). 

Moreover, the ambiguity reported is between two deprecated methods. So basically we won't really care.

I'd leave your code as it is—you already did the right thing.

PR #188, opened 2020-12-18
Add a fast toArray to ArraySet.

techsy730:
Add a fast `toArray` to `ArraySet`

Also slightly expand the unit tests for object based `toArray`s

### DIFF
--- a/drv/ArraySet.drv
+++ b/drv/ArraySet.drv
@@ -369,6 +369,41 @@ public class ARRAY_SET KEY_GENERIC extends ABSTRACT_SET KEY_GENERIC implements j
 
 	@Override
 	public boolean isEmpty() { return size == 0; }
+	
+#if KEYS_PRIMITIVE
+	@Override
+	public KEY_TYPE[] TO_KEY_ARRAY() {
+		return java.util.Arrays.copyOf(a, size);
+	}
+
+	@Override
+	public KEY_TYPE[] toArray(KEY_TYPE a[]) {
+		if (a == null || a.length < size) a = new KEY_TYPE[size];
+		System.arraycopy(this.a, 0, a, 0, size);
+		return a;
+	}
+#else // KEYS_REFERENCE
+	@Override
+	public Object[] toArray() {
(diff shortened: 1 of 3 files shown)


vigna:
Hey, toArray() with a primitive-type array argument is deprecated upstream. I really think you have warnings disabled for deprecated methods, or you should see this. My suggestion is that of not implementing it at all (the docs say it will be removed in a future release).

techsy730:
Oops, I thought I overrode the right primitive toArray. I guess not.

And yep, "Signal overriding or implementing deprecated method" was unchecked for me. Oops

vigna:
Yeah, probably we will get rid of it in 9.0.0. One day.

techsy730:
Actually, it is methods like `toIntArray(int[])` that are deprecated. It says `toArray(int[])` is the preferred one.

> ```java
	/** Returns a primitive type array containing the items of this collection.
	 *
	 * <p>Note that, contrarily to {@link Collection#toArray(Object[])}, this
	 * methods just writes all elements of this collection: no special
	 * value will be added after the last one.
	 *
	 * @param a if this array is big enough, it will be used to store this collection.
	 * @return a primitive type array containing the items of this collection.
	 * @see Collection#toArray(Object[])
	 * @deprecated Please use {@code toArray()} instead&mdash;this method is redundant and will be removed in the future.
	 */
	@Deprecated
	default KEY_TYPE[] TO_KEY_ARRAY(KEY_TYPE a[]) {
		return toArray(a);
	}

	/** Returns an array containing all of the elements in this collection; the runtime type of the returned array is that of the specified array.
	 *
	 * <p>Note that, contrarily to {@link Collection#toArray(Object[])}, this
	 * methods just writes all elements of this collection: no special
	 * value will be added after the last one.
	 *
	 * @param a if this array is big enough, it will be used to store this collection.
	 * @return a primitive type array containing the items of this collection.
	 * @see Collection#toArray(Object[])
	 */
	KEY_TYPE[] toArray(KEY_TYPE a[]);
```

vigna:
Oops, I remembered incorrectly. Well, at least now you have the warning on :).

PR #189, opened 2020-12-18
Add some @apiNotes to places they should have been.

techsy730:
Also, one unrelated change to List.sort it wouldn't always mention that `null` comparator means natural order for the Object based Lists.


vigna:
Seems fair. There are a few "strengthen" around that might lead you to similar fixes, if you wanna do that.

Issue #193, opened 2020-12-20
Parallel static utilities should use current ForkJoinPool if already in one


techsy730:
The parallel static utility methods should use the current ForkJoinPool if already running in one. This gives the user the option to use a ForkJoinPool they are maintaining instead of being forced to use the common one (which they may not want for prioritization reasons or similar). This is done by calling the static method in a task itself that is submitted to the pool they want to use.
This is also the only way you can control which pool is used for the JVM parallel operations (like `java.util.Arrays.parallelSort`)

[`ForkJoinTask.getPool()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinTask.html#getPool--), which is actually a static method despite the name, can be used to tell if currently in a ForkJoinPool or not.

Stretch goal would be to have overloads that let the user give their own ExecutorService they want to use (since the above being the standard library's only method of controlling this is a bit dumb TBH)

vigna:
So the best practice would be trying getPool() and then commomPool() in the former is null?

techsy730:
That is what I had in mind. We may want a not type specific static helper method for this.

I should not the "traditional" way is to submit the first task with `ForkJoinTask.fork()`, which will do this behind the scenes. But that only works if the first task is a `ForkJoinTask`.
Plus I honestly like being more explicit, and this makes it easier to be more flexible with later API evolution (like introducing a variant that takes an arbitrary ExecutorService).

Speiger:
My Solution to that was do not directly use ForkJoinPool itself. Have a Helper class where the CommonPool is set by default, but you can replace it and all sorting "Multithreading" tasks just go to the helper with their task that they want to execute.
Since invoke also automatically awaits the finish of the task this should be fine.

On top of that you can set a custom ForkJoinPool in the helper and if you try to set a "null" you just set the commonpool instead.
Straight out of SanityChecks
```java
	private static ForkJoinPool WORK_POOL = ForkJoinPool.commonPool();
	/**
	 * Returns if the current thread-pool can handle multi-threading tasks
	 * @return true if the threadcount is bigger the 1
	 */
	public static boolean canParallelTask() {
		return WORK_POOL.getParallelism() > 1;
	}
	
	/**
	 * Helper method to start a ForkJoinTask. This method will await the finalization of said task
	 * @param task the Task to invoke
	 */
	public static <T> void invokeTask(ForkJoinTask<T> task) {
		WORK_POOL.invoke(task);
	}
	
	/**
[...]

PR #194, opened 2020-12-21
Make fastutil use the current ForkJoinPool if already in one

techsy730:
Make fastutil use the current ForkJoinPool if already in one

Removes the `shutdown()` calls as we don't want to be shutting down users' Executors (we were getting away with it earlier because the common pool swallows shutdown requests).

Fixes https://github.com/vigna/fastutil/issues/193


src/it/unimi/dsi/fastutil/Threading.java
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Returns the current default executor that will be used by Fastutil's parallel operations not already in a {@link ForkJoinPool}.
+	 */
+	public static ForkJoinPool getCurrentDefaultFastutilExecutor() {
+		return defaultExecutor;
+	}
+
+	/** Set the current default executor that will be used by Fastutil's parallel operations not already in a {@link ForkJoinPool}.
+	 * 
+	 * <p>Note that this is static for the whole program. To override the pool to use for just a
```
vigna, line 70: I'm not sure this is s good idea. If in an application several different methods use fastutil, they could stomp on each other's feet. They could set the default execute at each call, but it seems overkill.

techsy730: Fair enough. Want to just leave the default pool hardcoded to common pool?

vigna: Yes, I'd just copy-paste where needed something like getPool() != null  ? getPool() : commonPool().

BTW thanks for fixing this, I'm really not an expert on Java parallelism.

techsy730: You mean not even have the Threading class at all? Just inline conditional?

vigna: Presently to have a class with a single method that is a single conditional seems overkill. I've thought about our places to place the utility methods (e.g., Arrays) but nothing seems to be right. Suggestions are welcome...

techsy730: How about `it.unimi.dsi.fastutil.Arrays`? That is the handwritten, non-typespecific one.
It is a little out of place for a class of that name but currently only Arrays classes have parallel operations.

If not that, it would only be two lines of code to inline into every location. Not that bad.

vigna: Yeah, I thought of Arrays, but still it's not the right place. It's two lines.

techsy730:
Ok, got rid of Threading and just put the static utility method in each class. A bit unfortunate, but any optimizer worth anything should be able to see these methods are tiny and inline them in short order.

vigna:
Can I merge this PR?

vigna:
OK, I'll wait tomorrow ;).

techsy730:
I think I am good now, but running unit tests just to be sure.

techsy730:
Looks like we are good. That last force-push was just to remove my local .gitignore that wasn't supposed to be committed.

PR #196, opened 2020-12-22
Make the email a little cleaner in @author tags.

techsy730:
It was hard to read escaped angle brackets.

Purely vanity I know, but it really did look ugly in source even though it looked fine in the HTML files.


vigna:
Angular brackets are standard for email addresses—they are also used in the CHANGES file. I'd live it as it is.

Issue #197, opened 2020-12-22
Resolve ambiguous overload warnings when using lambdas


techsy730:
An umbrella bug for discussion of overloads that can cause ambiguous calls if used with a lambda, making using them tricky.

The primary cause of this in fastutil is are a pair of Overloads that take a type-specific and an object based functor type.

Examples include (list will be updated as they are found)
* `Collection.removeIf`
* `Iterators.all`
* `Iterators.any`

vigna:
I have just checked in type-specific predicates and binary operators to help with this. 

I think the path you set up should be followed uniformly: there's a type-specific class which extends the type-generic version and the JDK type-specific version, possibly by widening. Then everything is deprecated and has a default implementation, except for the fastutil type-specific call.

Type-specific callers must implement and deprecate the generic call, as usual, and must implement and deprecate the JDK type-specific call, if it is available, so lambdas are not ambiguous.

techsy730:
OK, I have started with fixing `removeIf` by doing the disambiguation override thing that `forEach` and `forEachRemaining` does.

techsy730:
Do you happen to remember if the widening `removeIf` (like `removeIf(IntPredicate)` on a `ShortCollection`) was present in 8.4.0?
If not, we can just chuck the method out and have only the Object and the type-specific predicate overloads on shorter type collections.
If so, we should keep it around and deprecated it.

vigna:
Before we start to bang heavily on this, let us discuss for a second why we want to support JDK's primitive operators. The main reason I see is that there might be ready-made implementations of that interface that are actually useful. Am I right?

vigna:
The next question is: why do we want widened apply(int) in, say, CharConsumer? It is not useful for a lambda, and I don't see why there should be so many ready-made character-based consumers with an integer interface. But I might be wrong. And indeed the cost of having the widened apply() is fairly low.

techsy730:
For API compatibility if nothing else. If we just straight remove the Java library primitive predicate operations, then that will cause a NoSuchMethodError on already compiled code already calling the old one (as the overload is determined at compile time).
removeIf(IntPredicate) is already released API in the shorter data type collections.
For the shortened data types though I agree we should at least deprecate them.

For IntPredicate, LongPredicate, and DoublePredicate, IMO they don't really value add over the Java library's version and are mostly here for consistency. So I see no reason not to make the Java library's overloads the "primary" ones, and just have our versions extend the library's versions.

vigna:
[...]

PR #198, opened 2020-12-22
Deprecate the fact that shorted functors extend widened ones.

techsy730:
So deprecate the fact that, for example, ByteConsumer extends java.util.function.IntConsumer.

This fact has lead to some need some unfortunate extra overloads to avoid ambiguity.
These overloads were already marked deprecated, so when we remove the widened
superinterface we should also remove these methods.

Example of a deprecated method that should be co-removed is
ByteIterable.forEach(java.util.function.IntConsumer).

### DIFF
--- a/drv/Consumer.drv
+++ b/drv/Consumer.drv
@@ -20,13 +20,36 @@ package PACKAGE;
 import java.util.Objects;
 import java.util.function.Consumer;
 
+#if KEY_CLASS_Float
+/** A type-specific {@link Consumer}; provides methods to consume a primitive type both as object
+ * and as primitive.
+ *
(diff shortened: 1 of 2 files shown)


vigna:
I'm confused. Why deprecate? This stuff has never been distributed AFAIK.

techsy730:
For Predicate, you are right. We could just make the change.
But the Consumers were introduced in 8.0.0 it seems.

vigna:
Ok. At the same time, as I was discussing in #197 ,  predicates and consumers are exactly the case in which implementing the JDK interface,  even if the type does not match, makes perfect sense. Say you have a LongConsumer computing statistics on a stream: it is natural to use it on a CharStream. There is no unsafe type conversion. The situation is very different with operators.

PR #199, opened 2020-12-22
Fix removeIf ambiguity

techsy730:
The `removeIf` method had an ambiguity if you tried to use it with lambdas, requiring you to cast to a functor type yourself.

This change fixes it so it chooses a proper primitive Predicate.

It basically applies what was done to `forEach` and `forEachRemaining` to `removeIf`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,16 @@
 
 - New immutable array-based lists and associated List.of() methods.
 
+- Type specific versions of Predicate have been added.
+  Methods that take predicates (mainly Collection.removeIf) have been
+  updated to have an overload that takes that.
+  A consequence of this change is that removeIf(IntPredicate) on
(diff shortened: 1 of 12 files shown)


vigna:
OK, but unless I'm completely confused the logic here is opposite: in Iterator.drv, if there is a JDK-primitive implementation the fastutil-primitive implementation just delegates to that with a cast. Here the logic is the opposite: if there is a JDK-primitive implementation, it delegates to the fastutil-primitive implementation. Is there a technical reason for this difference?

vigna:
So in the JDK-primitive case there would always be a delegation in case we use a lambda. Wouldn't be more appealing to have removeIf(KEY_PREDICATE) contain the actual implementation and removeIf(JDK_PRIMITIVE_PREDICATE) to use a test / casting as in the non-JDK-primitive case? Why the different treatment?

Issue #201, opened 2020-12-24
Reorganize assertions, defines, and relationship between JDK's and fastutil's primitive-based functional interfaces


vigna:
This is an umbrella issue for discussing two problems that are becoming very visible in these days in the codebase:

- The overlap of different people implementing different conditional features has led to a proliferation of complex assertions—the semantics of which, however, is sometime much simpler than one would expect (e.g., it's just a matter of being int/long/double or any other type). All contributors added stuff to gencsources.sh, resulting in a large collection of #define's without any comment. Ten years ago that would have been OK, but with the amount of subtly convoluted conditional compilation we have now this is no longer acceptable. It has become very difficult (for me) to understand which code is being generated. There must be some documentation for the #define's, and clearly named assertions.

- The strategy for implementing functional interfaces or callers of functional interfaces in the presence of multiple possible implementations (e.g., object-based, JDK-primitive, or fastutil-primitive) is not uniform throughout the codebase. I find myself thinking most of the time "shouldn't this be implemented the other way around?". This is not healthy (oh, of course maybe I'm just old). Once again, ten years ago there were so few cases that looking around in other parts of the code was sufficient to understand what was the best practice, but this is no longer the case.

For the first part, it is mostly a matter of restructuring the file and commenting.

For the second part, I must write somewhere what is the preferred design. I'll keep updating this text, and I invite everybody working on fastutil to comment below.

Now, given the premises, there are things we cannot change:

- We want to benefit from code improvements upstream. This means that anything primitive that has been implemented in the JDK must not be overridden. For example, IntIerator.forEachRemaining() has primitive implementations in the JDK, so we should use those (possibly all of them—also the ones accepting object-based consumers).

- Lambdas will be fastutil-primitive. This is unavoidable because fastutil's types join the object and primitive JDK world, and this is the only possibility to avoid ambiguity while continuing fastutil's tradition of adding primitive-based operation without eliminating the object-based ones. As already implemented by @techsy730  in a number of cases, for methods accepting functional interfaces there must the usual deprecated object-based one, a fastutil-primitive one, and a JDK-primitive one if it is possible. We have by now functions (unary operators), predicates, consumers and binary operators, which should cover everything type-specific presently (except of course for two-argument functions in which the three types involved are not all the same, as in Map.computeIf*()).

The solution chosen is as follows (at present time, for Consumer/Predicate-related methods):

[...]

PR #203, opened 2020-12-29
Simplify forEach, forEachRemaining, and removeIf conditionals

techsy730:
Also adjusts the conditional of Iterable.drv, Iterator.drv, and Collection.drv to be in line with the style guidelines in https://github.com/vigna/fastutil/issues/201

### DIFF
--- a/drv/Collection.drv
+++ b/drv/Collection.drv
@@ -290,6 +290,9 @@ public interface COLLECTION KEY_GENERIC extends Collection<KEY_GENERIC_CLASS>, K
 	 *        removed.
 	 * @see Collection#removeIf(java.util.function.Predicate)
 	 * @return {@code true} if any elements were removed.
+	 * @apiNote Implementing classes that want to implement an optimized {@code removeIf} should
+	 *   generally override this method, and take the default implementation of the other overloads
+	 *   which will delegate to this method (after proper conversions).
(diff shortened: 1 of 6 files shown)


vigna:
OK. I see how you managed the three-legged thing and it's probably better than how I'm doing it in Consumer/Predicate, so I might go in the same direction: a macro for the type of the argument of the right implementation method, and a conditional to choose between delegation by casting or by new lambda. In this way appropriate comments can be added.

Minor comments:

- please refrain from using the <br> tag as it is deprecated as non-structural. A <p> is fine there;
- as in Consumer/Producer, I would not deprecate the method taking a widened Consumer or Predicate because one might have a ready-made one. I would just add an @implNote (see Consumer) explaining the inefficiency.

vigna:
I would also suggest adding tests for all types of overloads. I've done it for removeIf in shorts and ints. It is important not for the semantics, but to be sure that everything compiles as expected (e.g., no ambiguitity).

techsy730:
Ack on the tests; working on those now

techsy730:
No problem. I have a bad habit of being overly verbose to specify exactly what I mean, when there is really no risk one would get the wrong meaning from a simpler statement.
So helping me shorten my qualifiers is helpful. :)

PR #210, opened 2021-01-02
Add identity() to UnaryOperator

techsy730:
Adds the static method `identity` to UnaryOperator
For numeric primitives, adds `negation`, performing arithmetic negation.
For `BooleanUnaryOperator`, adds `negation`, performing logical negation.

For `BooleanPredicate`, it also adds those methods, as it also takes a boolean and returns a boolean, thus they make sense there too.

### DIFF
--- a/drv/Predicate.drv
+++ b/drv/Predicate.drv
@@ -36,7 +36,7 @@ import java.util.function.Predicate;
 #if ! KEY_CLASS_Boolean
 public interface KEY_PREDICATE KEY_GENERIC extends Predicate<KEY_GENERIC_CLASS>, JDK_PRIMITIVE_PREDICATE {
 #else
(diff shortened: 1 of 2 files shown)


vigna:
I'm a bit concerned by these semantic mixups. More than anything because each of these concepts has a different method naming. That's why IntUnaryOperator is not a Int2In2Function. I think we should keep the conceptual separation of the JDK unless there's some cogent reason.

techsy730:
Fair enough. How about I remove BooleanPredicate extending BooleanUnaryOperator, but keep the new identity and negation methods?

techsy730:
Ok done. `BooleanPredicate` no longer extends `BooleanUnaryOperator`

vigna:
Thanks, I forgot the identity method.

At this point was could as well add numeric negation to all other types except for char...

techsy730:
Ok, added a `negation` for numeric types.

Issue #200, opened 2020-12-23
Publish an image only using the basic primitives


incaseoftrouble:
If I recall correctly, I already added support for building a reduced jar ("fastutil-min"). It might make sense to actually publish this.

Reasoning: Asides from memory usage due to array packing there is, to my knowledge, no reason to use any key/value types except `int`, `long` and `double`, as essentially all operations are only defined for these types and operations on the other "primitives" even incur a performance penalty due to conversion. One could discuss whether `boolean` types could be included (`XY2BooleanMap` and `BooleanList` have some merit I guess). Advantage are smaller image (not that important), less clutter (autocompletion isn't spammed with 100 different types) and faster to work with (indexing fastutil does take a while).

Maybe the "full" fastutil image could even depend on the "min" image, however I guess this is a bit tricky to work with.

EDIT: `float`s are also natively supported by the JVM, however they come with many other pitfalls, so I still think that they should not be included in the core image

vigna:
That's a very good idea. Let's work it out. fastutil-core and fastutil-extra out something like that.

vigna:
I guess choosing int/long/double for fastutil-min makes a lot of sense because these are the types that get special treatment in the JDK.

vigna:
OK, at this point NO_SMALL_TYPES generates a smaller fastutil (1/3 of the classes). To avoid conditional compilation (i.e., to keep only conditional _inclusion_) some classes, such as Consumers or Functions, must be included for all types. But we gain a lot of space because of all missing map implementation.

If you wanna have look that'd be great—it's possible that still some class can be removed from the current set. The plan would be to release a "fastutil-core" and a "fastutil-extra" (depending on the first).

incaseoftrouble:
Is that on master?

vigna:
Yes

incaseoftrouble:
Ok, I'll have a look tomorrow / Wed

incaseoftrouble:
So two small things:
 * README should probably mention the small types switch
 * Maybe `NO_SMALL_TYPES` -> `ONLY_BASIC_TYPES` or something like that (positive switch)

[...]

PR #216, opened 2021-01-06
Core Types only compilation

incaseoftrouble:
See #200

### DIFF
--- a/src/it/unimi/dsi/fastutil/io/FastMultiByteArrayInputStream.java
+++ b/src/it/unimi/dsi/fastutil/io/FastMultiByteArrayInputStream.java
@@ -76,14 +76,28 @@ public class FastMultiByteArrayInputStream extends MeasurableInputStream impleme
 		for(int i = 0; i < array.length - 1; i++) {
 			array[i] = new byte[size >= SLICE_SIZE ? SLICE_SIZE : (int)size];
 			// It is important *not* to use is.read() directly because of bug #6478546
-			if (BinIO.loadBytes(is, array[i]) != array[i].length) throw new EOFException();
+			if (read(is, array[i]) != array[i].length) throw new EOFException();
 			size -= array[i].length;
 		}
 
 		current = array[0];
 	}
 
-	/** Creates a new multi-array input stream sharing the backing arrays of another multi-array input stream.
+  private static long read(final InputStream is, final byte[] a) throws IOException {
+    if (a.length == 0) return 0L;
+
+    int read = 0, result;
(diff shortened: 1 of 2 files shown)


src/it/unimi/dsi/fastutil/io/FastMultiByteArrayInputStream.java
```diff
@@ -76,14 +76,28 @@ public FastMultiByteArrayInputStream(final InputStream is, long size) throws IOE
 		for(int i = 0; i < array.length - 1; i++) {
 			array[i] = new byte[size >= SLICE_SIZE ? SLICE_SIZE : (int)size];
 			// It is important *not* to use is.read() directly because of bug #6478546
-			if (BinIO.loadBytes(is, array[i]) != array[i].length) throw new EOFException();
+			if (read(is, array[i]) != array[i].length) throw new EOFException();
```
vigna, line 79: I'm not sure this can be done. Bug #6478546 is still around in old JVMs. We might consider copying the code, but we cannot use just read().

vigna:
That seems to be fine. I'm just thinking whether we are doing this right. In the sense that if the purpose is to have a split distribution, more a conditional compilation we need build.xml to generate three different jars.

incaseoftrouble:
I have no idea about ant / maven though, I'm gradle-only :-D

Indeed, right now it would be three steps (something like `ONLY_BASIC_TYPES=1 make clean sources jar`, copy the jar, `IO_TYPES=1 make clean sources jar`, copy the jar, ...). However I am not sure how easy it is to define that in the build.xml. One could generate the files to different folders (`src/core/`, `src/io`, `src/full` or something like that). This would make building and IDE integration easy but may require some fiddling in the makefile (which I probably know too little about)

vigna:
So, I tried this and I get 1902 classes with `make -s clean sources MINIMAL_TYPES=1`. I don' think that was your intention, was it?

incaseoftrouble:
Huh.

```
make clean
MINIMAL_TYPES=1 make sources
find src/ -name "*.java" -print | wc -l
```
outputs `889` for me

vigna:
Actually, it's that the target clean needs fixing—$(TYPE_ALL) instead of $(TYPE). I was 

```
make -s clean sources MINIMAL_TYPES=1
```

and that was not cleaning the non-minimal types.

incaseoftrouble:
Oh oops, I missed that

vigna:
That has been like that forever, I just discovered the problem :).

PR #217, opened 2021-01-12
Minor touchups

techsy730:
Minor touchups.

* Add Javadoc to Iterators.any/all/indexOf and Spliterators.forEachMatching
* Undeprecate the JDK primitive overloads of any/all/indexOf. Now the overhead is documented as an `@implNote` to conform to the other similar methods in fastutil.
* Make an overload of `Spliterators.forEachMatching` that accept the JDK primitive operators.
* Make `List.replaceAll(JDK_PRIMITIVE_UNARY_OPERATOR)` forward to `List.replaceAll(KEY_UNARY_OPERATOR)` if possible, because, for example, `ByteUnaryOperator` is a subinterface of `java.util.function.IntUnaryOperator`.
* Convert iterator accepting methods of `Iterators` that don't need the fastutil specific methods to use `STD_KEY_ITERATOR`, which allows the regular `java.util.Iterator` to be accepted for the `ObjectIterators` class.
* Convert spliterator accepting methods of `Spliterators` that don't need to fastutil specific methods to use the new `STD_KEY_SPLITERATOR`, which allows the regular `java.util.Spliterator` to be accepted for the `ObjectSpliterators` class.

### DIFF
--- a/drv/Iterators.drv
+++ b/drv/Iterators.drv
@@ -832,7 +832,7 @@ public final class ITERATORS {
 	 * Returned whether any element satisfies the given predicate.
 	 * Short circuit evaluation is performed; the first {@code true} from the predicate terminates the loop.
 	 */
-	public static KEY_GENERIC boolean any(final KEY_ITERATOR KEY_GENERIC iterator, final METHOD_ARG_PREDICATE predicate) {
+	public static KEY_GENERIC boolean any(final STD_KEY_ITERATOR KEY_GENERIC iterator, final METHOD_ARG_PREDICATE predicate) {
 		return indexOf(iterator, predicate) != -1;
 	}
 #if KEYS_BYTE_CHAR_SHORT_FLOAT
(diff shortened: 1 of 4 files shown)


drv/Iterators.drv
```diff
@@ -828,19 +828,29 @@ public final class ITERATORS {
 		return new ListIteratorWrapper KEY_GENERIC_DIAMOND(i);
 	}
 	
+	/**
+	 * Returned whether any element satisfies the given predicate.
+	 * Short circuit evaluation is performed; the first {@code true} from the predicate terminates the loop.
+	 */
```
vigna, line 834: One observation: for all methods that require an iterator and do not use the additional methods of Iterator.drv, the right macro is STD_KEY_ITERATOR, which expands to Iterator for objects. That makes all the methods more generic at no cost. I would replace everywhere.

drv/List.drv
```diff
@@ -299,7 +299,7 @@ public interface LIST KEY_GENERIC extends List<KEY_GENERIC_CLASS>, COLLECTION KE
 	 *   lambda to perform widening and narrowing casts. Please use the type-specific overload to avoid this overhead.
 	 */
 	default void replaceAll(final JDK_PRIMITIVE_UNARY_OPERATOR operator) {
-		replaceAll(x -> KEY_NARROWING(operator.JDK_PRIMITIVE_KEY_APPLY(x)));
+		replaceAll(operator instanceof METHOD_ARG_KEY_UNARY_OPERATOR ? (METHOD_ARG_KEY_UNARY_OPERATOR) operator : x -> KEY_NARROWING(operator.JDK_PRIMITIVE_KEY_APPLY(x)));
```
vigna, line 302: Argh you're right I forgot about the check...

drv/Iterators.drv
vigna: And I believe there is at least one method in Spliterators.drv that might benefit from that approach.

techsy730: Both are done

vigna:
Amazing work, thanks.

PR #220, opened 2021-01-15
Further document the spliterator methods in interfaces.

techsy730:
Also pull the RandomAccess check for the default implementation of
spliterator() from AbstractList into List, to match up with what java.util.List does.

The detail is rather excruciating, but spliterator is kind of a nuanced API like that requiring that level of detail, at least in the interfaces it does.


techsy730:
Hang on, I just remembered I need to do it for BigList too.

PR #225, opened 2021-01-23
Fix a rather critical bug where ImmutableList.equals would return true erroneosly

techsy730:
Would happen to return `true` if any ImmutableList was given, equal contents or not.

### DIFF
--- a/drv/ImmutableList.drv
+++ b/drv/ImmutableList.drv
@@ -451,10 +451,10 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
 
 #if KEY_CLASS_Reference
 		while(s-- !=  0) if (a1[s] != a2[s]) return false;
-#else
-		java.util.Arrays.equals(a1, a2);
-#endif
 		return true;
+#else
+		return java.util.Arrays.equals(a1, a2);
+#endif
 	}
 
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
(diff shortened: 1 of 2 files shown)


drv/ImmutableList.drv
```diff
@@ -451,10 +451,10 @@ public class IMMUTABLE_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implem
 
 #if KEY_CLASS_Reference
 		while(s-- !=  0) if (a1[s] != a2[s]) return false;
+		return true;
```
vigna, line 454: But this is wrong anyway—if keys are references we must use equals().

Said that, this is a wrong copy-and-paste from ArrayList.drv. I think returning Arrays.equals() if perfectly fine in all cases—the previous code could not assume that the two backing arrays had the same length.

techsy730: `KEY_CLASS_Reference` is the case where we are making identity based collection (that test based on being the exact same object, not just `equals`)
`KEYS_REFERENCE` is the one that covers both Object and the Reference variants.

I think maybe we need a macro, `KEYS_USE_REFERENCE_EQUALITY` or something, that is the same value as `KEY_CLASS_Reference` but makes it clearer we are targeting reference equality semantics exactly.

vigna: Conclusion: I shouldn't read code on a Saturday when the Friday night had been like yesterday. 🤷🏻‍♂️

PR #227, opened 2021-01-23
Replace KEY_CLASS_Reference with KEYS_USE_REFERENCE_EQUALITY

techsy730:
This is much clearer that we mean the types that use `==` for equality testing instead of `.equals`, and not confuse it with the `KEYS_REFERENCE` macro which means any types using Object backed members.

Also documents the gotcha of trying to use a Reference based and a non-Reference based collection together (e.g. you may get non symmetric `equals`)

Also does the same for VALUE.


vigna:
This is a very good idea. I always find myself trying to remember the difference between KEY_CLASS_Object and KEY_CLASS_Reference. This is so much better.

PR #232, opened 2021-02-01
Fix issues with Map view iterators and spliterators

techsy730:
* Fix issues where not all the ways to iterate `keySet` or `values` in a LinkedOpenHashMap would follow the proper order (instead falling back to regular OpenHashMap order).
* Make sure all OpenHashMap's (and their Linked and Custom cousins) views have proper iterators and spliterators, as well as proper `forEachRemaining` methods. This saves a lot of implicit casts by the generics supporting bridge methods the compiler inserts.
* Add unit tests for these
* Give some views of AbstractMap better `spliterator()` methods

### DIFF
--- a/drv/AbstractMap.drv
+++ b/drv/AbstractMap.drv
@@ -24,12 +24,17 @@ import VALUE_PACKAGE.VALUE_ABSTRACT_COLLECTION;
 import VALUE_PACKAGE.VALUE_ITERATOR;
 import VALUE_PACKAGE.VALUE_SPLITERATOR;
 import VALUE_PACKAGE.VALUE_SPLITERATORS;
+#if VALUES_BYTE_CHAR_SHORT_FLOAT || VALUE_CLASS_Boolean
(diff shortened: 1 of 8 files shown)


techsy730:
Actually, I need to revert those last two revisions. In both the in the JDK and FastUtils, no spliterator implementation backed by an iterator tries to use the iterator in multiple threads. So it is actually safe to use to use them as stream sources.

vigna:
Ooooook... looks like that part was only partially written. I know, the conditional code is a bit crazy in the hash-based classes. Let me know when this is read to merge as I want to release 8.5.2.

techsy730:
Ok, I think we are good now.
But do we have any tests for `AbstractMap`? As in a map type (possible test only) that doesn't implement more then the bare minimum of methods require to extend `AbstractMap`?

vigna:
Have a look at AbstractInt2IntMapTest.

vigna:
In general, look in the IDE for Abstract*Test*.

PR #234, opened 2021-02-01
Fix AbstractMap containing a stack overflow containsValue

techsy730:
If both the default `containsValue` and the default `values` methods are
inherited by an implementating class, they will get a `StackOverflowError`
on trying to test for a value being present.

This is because the default `values().contains` calls `containsValue`, but
the default `containsValue` calls `values().contains`.

Now make the default `containsValues` loop over the entry set.

Add a comment that you should probably override this is your `values()`
has a better implementation. Same also retroactively applied to `containsKey`
and `keySet()`.

Finally, retrofits `AbstractInt2IntMapTest` to extend`Int2IntMapGenericTest`, with a dummy testonly implementation that implements the bare minimum for a mutable Int2Int map. This is to test as many of the `AbstractInt2IntMap`'s default methods as possible. This gives it the set of tests we apply to all other `Int2IntMap`  impls.

### DIFF
--- a/drv/AbstractMap.drv
+++ b/drv/AbstractMap.drv
@@ -307,7 +307,7 @@ public abstract class ABSTRACT_MAP KEY_VALUE_GENERIC extends ABSTRACT_FUNCTION K
 				@Override
 				public KEY_SPLITERATOR KEY_GENERIC spliterator() {
 					return SPLITERATORS.asSpliterator(
-	 					iterator(), sizeOf(ABSTRACT_MAP.this), SPLITERATORS.SET_SPLITERATOR_CHARACTERISTICS);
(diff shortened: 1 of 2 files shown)


techsy730:
Hopefully not too late for 8.5.2.

If it is, I don't think it is a big enough issue for a 8.5.3 alone. It can just be rolled up into whatever other fixes do deserve it.

vigna:
No, we're definitely in time for 8.5.2.

techsy730:
Ok, I am going to leave this change as is most likely unless you spot something major.

vigna:
OK, I merged in and removed my commit about varargs (it's a mess—you get confusion between the constructor with a size argument and the constructor with one element). I had to do some github black sorcery—I just hope I didn't delete any of your merges.

PR #242, opened 2021-03-14
Fix toXWithExpected size overallocating in parallel stream.

### LINKED ISSUE #211: The `toXWithExpectedSize` methods pre-allocate a full size array for every thread.
When using, for example, `IntArrayList.toListWithExpectedSize` in a parallel stream, every thread will pre-allocate the size given.
While this doesn't break correctness, it is rather wasteful, especially if given a large size on a highly multithreaded system.

Only one thread (the one that will merge it all into the final result) needs the full array. The second thread only needs ~half of the given size preallocated (as it will only be handling about half the input). The 3rd and 4th only need a ~quarter of the size preallocated. Etc

For the `toXWithExpectedSize` collectors may want to replace the supplier function (the one that allocates each threads' mutable accumulator) with a stateful one that tracks how times it has been called and reduce how much it preallocates as the calls go up (to some floor, like 8 or something).
The "correct" way would be `preallocateSize / (floor(log_2(timesCalled)) + 1)`, but that may be too complicated to be worth it.
A simpler method that isn't quite as optimal but still gives a meaningful improvement to the current state is `preallocateSize / timesCalled`.

This assumes that the first thread is indeed the one that will merge together the final result. If it isn't, then all this gets thrown out the window and there is no safe way we can determine which threads get how much.

Alternatively, we can leave the methods as they are and just warn as an `@implNote` that these methods may be a bit wasteful with memory if used on parallel streams.

techsy730:
Fix toXWithExpected size overallocating in parallel stream.

This is accomplished by having the `Supplier` preallocate a `Collection`
of full expected size for the first collection allocation in the first thread,
1/2 expected size for the second, 1/3 for the third, etc.

Fixes #211

This reduces the minimum _memory_ allocated when using this in a parallel stream from
Ω(threads \* expectedSize) -> Ω(threads \* H_expectedSize) = Ω(threads \* ln(expectedSize)).
(The worst case memory usage is still O(threads \* expectedSize), but that requires some truly pathological splitting decisions by the backing `Spliterator`, and is a problem the non-sizing logic would have too)

Where Ω is the min asymptotically bounding function (Think O notation but for the min instead of max)
threads is the number of threads the `Stream` will use
expectedSize is the value of the expectedSize parameter given to `toXWithExpectedSize`
and H_n is the nth harmonic number, 1 + 1/2 + 1/3 + ... + 1/n
  For a more familiar complexity comparison, Θ(H_n) = Θ(ln(n)) (Θ is like O notation but bounded above and below)
  as the difference between H_n and ln(n) approaches a constant as n->∞ (the Euler–Mascheroni constant for those interested)

<sub><sub>...What are you talking about, I'm not just flexing math knowledge. <\_< &nbsp; >\_> </sub></sub>

It's certainly not perfect. The 4nd thread isn't assured to be the 4th to allocate for example. And not all Spliterators take a split-by-two approach. But it shouldn't be any worse in performance then not providing a size estimate at all (unless you radically overestimate)

### DIFF
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -40,7 +40,7 @@ import java.util.function.Consumer;
  * <em>never</em> made smaller (even on a {@link #clear()}). A family of
  * {@linkplain #trim() trimming methods} lets you control the size of the
  * backing array; this is particularly useful if you reuse instances of this class.
- * Range checks are equivalent to those of {@link java.util}'s classes, but
+ * Range checks are equivalent to those of {@code java.util}'s classes, but
  * they are delayed as much as possible. The backing array is exposed by the
  * {@link #elements()} method.
  *
@@ -340,7 +340,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		return wrap(a, a.length);
 	}
 
-	/** Creates a new empty array list. 
+	/** Creates a new empty array list.
 	 *
 	 * @return a new empty array list.
 	 */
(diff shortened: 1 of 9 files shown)


techsy730:
Ok, I think I am done for now. Core logic seems ok, only Javadoc changes may be possible (but no need to block PR over that)

vigna:
> Where Ω is the min asymptotically bounding function (Think O notation but for the min instead of max)

Wow, never heard of it. 😂

PR #256, opened 2021-06-24
Add collectors for Maps

shawjef3:
This adds collectors for `Stream` to fastutil Maps.

Future improvements would be to investigate supporting primitive streams (e.g. `IntStream`) and improving the type signature for collectors where char is the key or value. When `char` is the key, the functions are for `int`. When `char` is the value, the functions are for `Object`.

### DIFF
--- a/drv/Maps.drv
+++ b/drv/Maps.drv
@@ -37,6 +37,11 @@ import java.util.Map;
 import java.util.function.Consumer;
 import PACKAGE.MAP.FastEntrySet;
 
+import java.util.function.BiConsumer;
+import java.util.function.BinaryOperator;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+
 /** A class providing static methods and objects that do useful things with type-specific maps.
  *
  * @see java.util.Collections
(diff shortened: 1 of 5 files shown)


vigna:
I won't be able to look at this for a couple of weeks as I'll be travelling. Did you use some of the infrastructure for primitive functions (look for PRIMITIVE_FUNCTION in gencsources.sh)?

shawjef3:
I tried to use JDK_PRIMITIVE_FUNCTION, but ran into many places it didn't fit. For instance, in `Byte2BooleanMaps`, the collector needs `Object2ByteFunction<T>` for the keys, and `Predicate<T>` for the values. However, `JDK_PRIMITIVE_FUNCTION_APPLY` gives `IntPredicate`, which isn't applicable to the collector.

vigna:
Ahem, I know I'm late, but I'm on vacation now. I'll be back to work in full in a couple of weeks.

vigna:
So, the way we did that for other containers is quite different—see, e.g., toSet() in OpenHashSet.drv. It would be better to try to follow the same scheme.

PR #259, opened 2021-07-09
Fix FastBufferedOutputStream.close() exception

stachenov:
If the output is closed before `FastBufferedOutputStream` is closed, then `close()` would throw because `close()` calls `flush()`, and it would try to write the buffer, even if there's nothing to write, and zero-length `write()` on a closed stream throws in some `OutputStream` implementations.

This can happen if the output is a socket, for example, that is managed separately from the output stream.

Fix by not calling `write()` from `flush()` at all if there is nothing to write.

If the buffer is not empty, then it would still try to write it, and would throw, which is probably correct.

### DIFF
--- a/src/it/unimi/dsi/fastutil/io/FastBufferedOutputStream.java
+++ b/src/it/unimi/dsi/fastutil/io/FastBufferedOutputStream.java
@@ -125,6 +125,7 @@ public class FastBufferedOutputStream extends MeasurableOutputStream implements 
 	}
 
 	private void dumpBuffer(final boolean ifFull) throws IOException {
+		if (pos == 0) return; // nothing to dump
 		if (! ifFull || avail == 0) {
 			os.write(buffer, 0, pos);
 			pos = 0;
--- a/test/it/unimi/dsi/fastutil/io/FastBufferedOutputStreamTest.java
+++ b/test/it/unimi/dsi/fastutil/io/FastBufferedOutputStreamTest.java
@@ -23,7 +23,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.FilterOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.channels.FileChannel;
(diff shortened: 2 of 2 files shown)


src/it/unimi/dsi/fastutil/io/FastBufferedOutputStream.java
```diff
@@ -125,6 +125,8 @@ public FastBufferedOutputStream(final OutputStream os) {
 	}
 
 	private void dumpBuffer(final boolean ifFull) throws IOException {
+		if (pos == 0)
+			return; // nothing to dump
```
shawjef3, line 129: Braces are good even if it's just one line.

https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

stachenov: Fair enough. I normally don't add braces to simple early returns and other typical one-line cases (judging by how typical a case is rather than by the actual number of lines), and that code is already full of single-line `if`s, so I thought it was OK... But certainly more braces won't hurt! Fixed.

incaseoftrouble: first and foremost it should be consistent with the style guide of the project. If one-line-returns don't use braces, don't start adding this - it causes way more confusion.

btw: using a code-formatter (correcting the indent) "prevents" the goto-fail in the same way as adding braces :-) the problem there is not the braces in particular but breaking with the code style / not properly indicating the end of a block. If braces were essential, python code would arguably be inherently unsafe ;-)

stachenov: This particular project doesn't seem to have a style guide, though.

Looking at the rest of the code, not only it seems not to use braces, it also seems to place both `if` and its body on the same line, which I'm not a fan of, but I guess following the style is probably the best choice. And besides, in case of a single `return` at the start of the function, it doesn't really matter much.

Re-pushed again, this time following the style of the rest of the files.

vigna:
OK, thanks for noting this!

PR #276, opened 2022-06-11
Removed excessive init in HeapIndirectPriorityQueue constructor

hpple:
Hi!

It looks like the work is already done in super constructor


vigna:
Yep, you're right.

PR #281, opened 2022-11-03
faster `HashCommon.nextPowerOf2`

richardstartin:
The integer version of this method showed up in a profile, and I was curious enough to look at its implementation. This provides a slightly faster implementation of `HashCommon.nextPowerOf2` which takes advantage of `numberOfLeadingZeros` intrinsics.

I wrote a very simple benchmark to justify the change. While it shows that the existing code is hardly worth optimising, the new code is indeed faster:

```
Benchmark                     Mode  Cnt  Score   Error  Units
NextPowerOf2Benchmark.after   avgt    5  0.553 ± 0.007  ns/op
NextPowerOf2Benchmark.before  avgt    5  0.916 ± 0.002  ns/op
```

### DIFF
--- a/src/it/unimi/dsi/fastutil/HashCommon.java
+++ b/src/it/unimi/dsi/fastutil/HashCommon.java
@@ -157,13 +157,7 @@ public class HashCommon {
 	 * @return the least power of two greater than or equal to the specified value.
 	 */
 	public static int nextPowerOfTwo(int x) {
-		if (x == 0) return 1;
-		x--;
-		x |= x >> 1;
-		x |= x >> 2;
-		x |= x >> 4;
-		x |= x >> 8;
-		return (x | x >> 16) + 1;
+		return 1 << (32 - Integer.numberOfLeadingZeros(x - 1));
 	}
 
 	/** Returns the least power of two greater than or equal to the specified value.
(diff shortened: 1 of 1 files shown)


vigna:
Very interesting how it handles the case 0. OK, but, just out of curiosity, did you try to benchmark with GraalVM?

richardstartin:
> Very interesting how it handles the case 0. OK, but, just out of curiosity, did you try to benchmark with GraalVM?

I did not, but I can try that and let you know.

richardstartin:
I got roughly the same scores (JDK 17.0.5, OpenJDK 64-Bit Server VM, 17.0.5+8-jvmci-22.3-b08, on an M1):

```
Benchmark                     Mode  Cnt  Score   Error  Units
NextPowerOf2Benchmark.after   avgt    5  0.647 ± 0.109  ns/op
NextPowerOf2Benchmark.before  avgt    5  0.965 ± 0.002  ns/op
```

vigna:
OK thanks!

richardstartin:
Oh no! I thought this was an elegant solution (looking for a problem, of course) :(

vigna:
Why no? I just merged...

vigna:
Ah OK for some weird reason I did not merge properly.

PR #290, opened 2023-03-13
Fix bug in ImmutableSubList iterator

goldfeder:
listIterator() on a sublist needs to correct its index to add "from" to it.


vigna:
Ouch, that must have been there for a long time. Could you add a unit test for that in IntImmutableListTest? Thanks!

goldfeder:
Done!

vigna:
Thanks! If you want to be thanked in the CHANGES file let me know your real name.

goldfeder:
Sure, why not. I'm Chanoch Goldfeder. Thanks!

goldfeder:
Ugh I just realized I fixed this wrong. It fixes the iteration but breaks
other stuff. Correct fix coming in a few.

PR #291, opened 2023-03-13
Continue fixing previous bug

goldfeder:
I didn't correct all of the methods


vigna:
OK, would it be too much to ask to check with unit tests the ImmutableSublist code? 🙈

goldfeder:
Of course not, I should have done that. But I found a few more related
bugs. Fixing all and will send with tests.

goldfeder:
Ok, please take a look now. I fixed a bunch of bugs in the sublist listiterator (now with tests) and also fixed one bug in the regular list iterator (back() checked the wrong value for remaining). Since there were no tests for the regular list iterator, I added them there too, even though I only fixed the bug in back().

vigna:
Fantastic, thanks!

PR #294, opened 2023-04-02
Add ImmutableSet types

goldfeder:
This ought to be more or less correct (code compiles, tests pass, seems functional) but I'd appreciate your feedback!

I think I've mostly figured out how the drivers work, but I decided to start with only the immutable sets, which are relatively simple, before doing the map types.

### DIFF
--- a/drv/ImmutableSet.drv
+++ b/drv/ImmutableSet.drv
@@ -1,0 +1,149 @@
+/*
+ * Copyright (C) 2002-2023 Sebastiano Vigna
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+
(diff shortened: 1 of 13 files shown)


vigna:
Wow, you totally got it without my help. Kudos, as sometimes when I don't work on it for a while *I* don't get it. 😂

The only things I would change are:
- Tabs in fastutil are actual tabs (currently in the code an indentation is two spaces). But that's very easy to fix.
- Probably we need a bit more documentation explaining in detail what is the difference with an immutable wrapper method from Sets, so people can choose what to use, as we cannot assume they are familiar with what Guava does. In theory, it would be great to have a `@see` in the methods of corresponding unmodifiable containers, but unfortunately one of the limitations of the method I'm using is that we don't have replacements in comments. Usually I circumvent this using some wordy description—I'll think of something.
- The license blurb should be at the start of test files, too.

goldfeder:
Thanks! It was pretty logical and I managed with a little trial and error.

I will make the requested changes (although I could use help with the docs)
but it will probably wait a week because we are on holidays here in Israel
and I am not going to be at a computer much.

After this looks good I’ll send you the maps code, and also some other
random fixes and utilities that you may (or may not) like. I did a bunch of
stuff with strengthening Stream to work better with fastutil that I’d love
to show you.

goldfeder:
Oh my goodness.

I switched jobs months and months ago and also moved and a bunch of other
stuff and in the process I totally forgot about this. Months later I
remembered and I was… embarrassed to reach out after having disappeared for
so long. And then after that I sort of put it in the back of my mind and
never contacted you.

I literally have the code ready to go. Can I finish this off? If you hadn’t
closed this I probably never would have remembered but if you are still
willing I can knock this out for real this week and have it submitted.

PR #296, opened 2023-04-20
Introduced `FixedArrayFIFOQueue`

catap:
This is a copy of `ArrayFIFOQueue` which has fixed capacity.

I've also refactor a bit `PriorityQueue` interface to introduce a clear way to use exception free queue.


catap:
@vigna I've heavy rework this PR to use `ArrayFIFOQueue` as a base version. I've removed all councurency things which was originally here because it is too complex and probably useless. Synchronization for few operation is near of nothing, and better to suggest users to make simpler solution when they need it.

PR #306, opened 2023-10-06
Make OpenHashSet a bit more concurrent friendly

catap:
Here a quite trivial reorganization of `OpenHashSet` which allows it to be used on that way:
```
    public static void main(String[] args) {
        LongSet original = new LongOpenHashSet();
        LongSet readOnly = LongSets.unmodifiable(original);
        LongSet writable = LongSets.synchronize(original);

        SplittableRandom random = new SplittableRandom();

        AtomicBoolean failure = new AtomicBoolean();
        new Thread(() -> {
            while (true) {
                try {
                    readOnly.contains(random.nextLong());
                } catch (Throwable th) {
                    failure.set(true);
                    break;
                }
            }
        }).start();
        new Thread(() -> {
            while (!failure.get()) {
                writable.add(random.nextLong());
            }
        }).start();
    }
```

The goal is crash at OutOfMemory, and not because of concurent access.

### DIFF
--- a/drv/OpenHashSet.drv
+++ b/drv/OpenHashSet.drv
@@ -971,7 +971,13 @@ public class OPEN_HASH_SET KEY_GENERIC extends ABSTRACT_SET KEY_GENERIC implemen
 		if (KEY_EQUALS_NULL(KEY_GENERIC_CAST k)) return containsNull;
 
 		KEY_GENERIC_TYPE curr;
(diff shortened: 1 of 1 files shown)


vigna:
Sorry, I really don't think it's a good idea.

catap:
np, it is hack, yes.

PR #314, opened 2024-02-06
ArrayList iterator back() bug fix, compute remaining correctly

barakugav:
The current implementation of ArrayList.iterator().back(int n) compute the number of remaining elements wrongly.
It compute `size-pos`, which is correct for the skip(int n) method, but wrong for the back(int n) method.


barakugav:
Together with https://github.com/vigna/fastutil/pull/291, these bugs are kind of limiting the use of the lists. Is there a plan for a new release soon?

vigna:
Good catch!

vigna:
Just done 8.5.13! Should be on sonatype soon.

barakugav:
Thanks!

PR #315, opened 2024-02-07
Immutable sub list iterator bug fix, ensureIndex is relative to sub range

barakugav:
ensureRestrictedIndex in abstract list compare to size(), 'from' should not be used.


vigna:
Omg. So many bugs. :(

PR #311, opened 2023-12-18
Cache class field ref into local

mouse0w0:

PR #316, opened 2024-02-12
Optimize ArrayList

mouse0w0:
In the previous pull request (#311), I optimized the array field access of some classes. Recently, with my deeper understanding of fastutil, I have discovered more optimizations and technical debts that need attention. However, resolving these issues is challenging and time-consuming, requiring contributions from the community and multiple pull requests. This pull request is the recent work I have done on ArrayList.

### DIFF
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -500,7 +500,8 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	@Override
 	public int indexOf(final KEY_TYPE k) {
 		final KEY_TYPE[] a = this.a;
-		for(int i = 0, max = size; i < max; i++) if (KEY_EQUALS(k, a[i])) return i;
+		final int size = this.size;
+		for(int i = 0; i < size; i++) if (KEY_EQUALS(k, a[i])) return i;
 		return -1;
 	}
 
@@ -632,7 +633,8 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		@Override
 		public void forEach(final METHOD_ARG_KEY_CONSUMER action) {
 			final KEY_GENERIC_TYPE[] a = ARRAY_LIST.this.a;
-			for (int i = from, max = to; i < max; ++i) {
+			final int to = this.to;
+			for (int i = from; i < to; ++i) {
 				action.accept(a[i]);
 			}
 		}
@@ -875,7 +877,8 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	@Override
 	public void forEach(final METHOD_ARG_KEY_CONSUMER action) {
 		final KEY_GENERIC_TYPE[] a = this.a;
-		for (int i = 0, max = size; i < max; ++i) {
+		final int size = this.size;
+		for (int i = 0; i < size; ++i) {
 			action.accept(a[i]);
 		}
 	}
@@ -1019,12 +1022,12 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 				@Override
 				public void forEachRemaining(final METHOD_ARG_KEY_CONSUMER action) {
 					final KEY_GENERIC_TYPE[] a = ARRAY_LIST.this.a;
-					final int max = ARRAY_LIST.this.size;
-					for (int i = pos; i < max; i++) {
+					final int size = ARRAY_LIST.this.size;
+					for (int i = pos; i < size; i++) {
 						action.accept(a[i]);
 					}
-					pos = max;
-					last = max - 1;
+					pos = size;
+					last = size - 1;
 				}
 				@Override
 				public int back(int n) {
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -524,6 +524,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 #if KEYS_REFERENCE
 		a[size] = null;
 #endif
+		assert size <= a.length;
 		return old;
 	}
 
@@ -532,6 +533,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		int index = indexOf(k);
 		if (index == -1) return false;
 		REMOVE_KEY(index);
+		assert size <= a.length;
 		return true;
 	}
 
@@ -541,6 +543,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		final KEY_GENERIC_TYPE[] a = this.a;
 		final KEY_GENERIC_TYPE old = a[index];
 		a[index] = k;
+		assert size <= a.length;
 		return old;
 	}
 
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -541,7 +541,6 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		final KEY_GENERIC_TYPE[] a = this.a;
 		final KEY_GENERIC_TYPE old = a[index];
 		a[index] = k;
-		assert size <= a.length;
 		return old;
 	}
 
@@ -607,6 +606,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		final KEY_GENERIC_TYPE[] t = KEY_GENERIC_ARRAY_CAST new KEY_TYPE[Math.max(n, size)];
 		System.arraycopy(a, 0, t, 0, size);
 		this.a = t;
+		assert size <= a.length;
 	}
 
 	private class SubList extends ABSTRACT_LIST.SUBLIST_RANDOM_ACCESS KEY_GENERIC {
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -663,10 +663,12 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 			@Override
 			public void forEachRemaining(final METHOD_ARG_KEY_CONSUMER action) {
 				final KEY_GENERIC_TYPE[] a = ARRAY_LIST.this.a;
-				final int max = to - from;
-				for (int i = pos; i < max; i++) {
+				final int from = this.from, to = this.to;
+				for (int i = from + pos; i < to; i++) {
 					action.accept(a[i]);
 				}
+
+				final int max = to - from;
 				pos = max;
 				lastReturned = max - 1;
 			}
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -449,11 +449,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		a = ARRAYS.ensureCapacity(a, capacity, size);
 #else
 		if (wrapped) a = ARRAYS.ensureCapacity(a, capacity, size);
-		else {
-			final Object[] t = new Object[capacity];
-			System.arraycopy(a, 0, t, 0, size);
-			a = (KEY_GENERIC_TYPE[])t;
-		}
+		else a = Arrays.copyOf(a, capacity);
 #endif
 		assert size <= a.length;
 	}
@@ -473,11 +469,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		a = ARRAYS.forceCapacity(a, capacity, size);
 #else
 		if (wrapped) a = ARRAYS.forceCapacity(a, capacity, size);
-		else {
-			final Object[] t = new Object[capacity];
-			System.arraycopy(a, 0, t, 0, size);
-			a = (KEY_GENERIC_TYPE[])t;
-		}
+		else a = Arrays.copyOf(a, capacity);
 #endif
 		assert size <= a.length;
 	}
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -442,7 +442,6 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 *
 	 * @param capacity the new minimum capacity for this array list.
 	 */
-	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	public void ensureCapacity(final int capacity) {
 		if (capacity <= a.length || (a == ARRAYS.DEFAULT_EMPTY_ARRAY && capacity <= DEFAULT_INITIAL_CAPACITY)) return;
 #if KEYS_PRIMITIVE
@@ -459,7 +458,6 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 *
 	 * @param capacity the new minimum capacity for this array list.
 	 */
-	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	private void grow(int capacity) {
 		if (capacity <= a.length) return;
 		if (a != ARRAYS.DEFAULT_EMPTY_ARRAY)
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -1210,15 +1210,15 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	public boolean equals(final ARRAY_LIST KEY_GENERIC l) {
 		// TODO When minimum version of Java becomes Java 9, use the Arrays.equals which takes bounds, which is vectorized.
 		if (l == this) return true;
-		int s = this.size;
-		if (s != l.size()) return false;
+		final int size = this.size;
+		if (size != l.size()) return false;
 		final KEY_GENERIC_TYPE[] a1 = a, a2 = l.a;
 		if (a1 == a2) return true;
 
 #if KEY_CLASS_Object
-		while(s-- !=  0) if (! java.util.Objects.equals(a1[s], a2[s])) return false;
+		for (int i = 0; i < size; i++) if (!java.util.Objects.equals(a1[i], a2[i])) return false;
 #else
-		while(s-- !=  0) if (a1[s] != a2[s]) return false;
+		for (int i = 0; i < size; i++) if (a1[i] != a2[i]) return false;
 #endif
 		return true;
 	}
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -665,7 +665,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 			@Override
 			public void forEachRemaining(final METHOD_ARG_KEY_CONSUMER action) {
 				final KEY_GENERIC_TYPE[] a = ARRAY_LIST.this.a;
-				final int from = this.from, to = this.to;
+				final int from = SubList.this.from, to = SubList.this.to;
 				for (int i = from + pos; i < to; i++) {
 					action.accept(a[i]);
 				}
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -828,15 +828,14 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 */
 	@Override
 	public void removeElements(final int from, final int to) {
+		final int size = this.size;
 		it.unimi.dsi.fastutil.Arrays.ensureFromTo(size, from, to);
 		final KEY_TYPE[] a = this.a;
 		System.arraycopy(a, to, a, from, size - to);
-		size -= (to - from);
 #if KEYS_REFERENCE
-		for (int i = from; i < to; i++) {
-			a[i] = null;
-		}
+		Arrays.fill(a, to, size, null);
 #endif
+		this.size -= (to - from);
 	}
 
 
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -117,7 +117,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	 * with {@link #wrap}.
 	 */
 	SUPPRESS_WARNINGS_KEY_UNCHECKED
-	private static final KEY_GENERIC KEY_GENERIC_TYPE[] copyArraySafe(KEY_GENERIC_TYPE[] a, int length) {
+	private static KEY_GENERIC KEY_GENERIC_TYPE[] copyArraySafe(KEY_GENERIC_TYPE[] a, int length) {
 		if (length == 0) return KEY_GENERIC_ARRAY_CAST ARRAYS.EMPTY_ARRAY;
 #if KEYS_PRIMITIVE
 		return java.util.Arrays.copyOf(a, length);
@@ -126,7 +126,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 #endif
 	}
 
-	private static final KEY_GENERIC KEY_GENERIC_TYPE[] copyArrayFromSafe(ARRAY_LIST KEY_GENERIC l) {
+	private static KEY_GENERIC KEY_GENERIC_TYPE[] copyArrayFromSafe(ARRAY_LIST KEY_GENERIC l) {
 		return copyArraySafe(l.a, l.size);
 	}
 
@@ -450,11 +450,9 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 #else
 		if (wrapped) a = ARRAYS.ensureCapacity(a, capacity, size);
 		else {
-			if (capacity > a.length) {
-				final Object[] t = new Object[capacity];
-				System.arraycopy(a, 0, t, 0, size);
-				a = (KEY_GENERIC_TYPE[])t;
-			}
+			final Object[] t = new Object[capacity];
+			System.arraycopy(a, 0, t, 0, size);
+			a = (KEY_GENERIC_TYPE[])t;
 		}
 #endif
 		assert size <= a.length;
@@ -488,6 +486,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	public void add(final int index, final KEY_GENERIC_TYPE k) {
 		ensureIndex(index);
 		grow(size + 1);
+		final KEY_GENERIC_TYPE[] a = this.a;
 		if (index != size) System.arraycopy(a, index, a, index + 1, size - index);
 		a[index] = k;
 		size++;
@@ -511,7 +510,7 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 	@Override
 	public int indexOf(final KEY_TYPE k) {
 		final KEY_TYPE[] a = this.a;
-		for(int i = 0; i < size; i++) if (KEY_EQUALS(k, a[i])) return i;
+		for(int i = 0, max = size; i < max; i++) if (KEY_EQUALS(k, a[i])) return i;
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


incaseoftrouble:
Did you profile these changes? (With a proper JMH setup)

I'm doubtful that these changes have any noticeable performance improvements (the JVM is incredibly good at figuring out these things). In particular, copying fields to local variables in my experience just adds clutter and doesn't yield anything. (Why I come to this conclusion: I also thought that it helps, after all, the compiler cannot prove the absence of concurrent modification. But even in really costly iterative methods this made practically no difference. I think this is partly due to such a local variable needing to be copied and *written* to memory, since, indeed, the compiler cannot prove that not copying the value will have no influence.)

Similarly, marking local variables as final has (hardly) any influence. To my knowledge, the only possibility is that the compiler can perform some static analysis (which the JIT would very likely figure out, too) - `final` is discarded from the byte-code.

mouse0w0:
I will conduct a benchmark for List for each. The benchmark code is as follows:
```java
@State(Scope.Thread)
@Warmup(iterations = 3)
@Measurement(iterations = 10)
@Fork(1)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class IterationBenchmark {
    @Param({"10", "1000", "100000"})
    private int n;

    private int[] a;

    @Setup
    public void setup() {
        int[] a = new int[n];
        for (int i = 0; i < n; i++) {
            a[i] = i;
        }
        this.a = a;
    }

    @Benchmark
    public void forEach(Blackhole blackhole) {
        for (int i = 0; i < n; i++) {
            blackhole.consume(a[i]);
        }
    }

    @Benchmark
    public void forEachCacheArray(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0; i < n; i++) {
            blackhole.consume(a[i]);
        }
    }

    @Benchmark
    public void forEachCacheSize(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0, max = n; i < max; i++) {
            blackhole.consume(a[i]);
        }
    }
}
```
The benchmark results are as follows:
```
Benchmark                                (n)  Mode  Cnt      Score     Error  Units
IterationBenchmark.forEach                10  avgt   10      7.003 ±   0.006  ns/op
IterationBenchmark.forEach              1000  avgt   10    656.301 ±   0.153  ns/op
IterationBenchmark.forEach            100000  avgt   10  65878.207 ± 424.070  ns/op
IterationBenchmark.forEachCacheArray      10  avgt   10      4.208 ±   0.004  ns/op
IterationBenchmark.forEachCacheArray    1000  avgt   10    330.056 ±   0.024  ns/op
IterationBenchmark.forEachCacheArray  100000  avgt   10  32982.049 ±   6.810  ns/op
IterationBenchmark.forEachCacheSize       10  avgt   10      2.805 ±   0.004  ns/op
IterationBenchmark.forEachCacheSize     1000  avgt   10    105.135 ±   0.013  ns/op
IterationBenchmark.forEachCacheSize   100000  avgt   10  13622.145 ±   9.760  ns/op
```
Based on the benchmark results, caching the array object and size of the list reduces iteration time by over 70% compared to not caching. Regarding the benchmark results, my opinions are as follows:

1. When the code attempts to get a field of the `this` object, it needs to execute two opcodes, `aload 0` and `getfield`. However, when getting a local variable, only one opcode is required, such as `iload 1`. Therefore, if the code needs to access a field value more than twice, the number of opcodes for accessing fields is greater than accessing local variables.

2. Local variables are stored in the stack rather than the heap, implying that they can be released from memory when the method returns.

3. Computers tend to store local variables in the CPU cache rather than memory, facilitating faster access.

4. The `getfield` opcode is relatively time-consuming, which to some extent affects performance.

Furthermore, using the final modifier to declare local variables is to conform to the existing code style of fastutil. Looking forward to your reply.

incaseoftrouble:
> Based on the benchmark results, caching the array object and size of the list reduces iteration time by over 70% compared to not caching. 

... on this benchmark that doesn't have dynamic dispatch etc. involved and is about forEach (I'm totally in favor of specializing `forEach`!)

I doubt that there will be any difference (there might even be a penalty) for `get` and `set` (i.e. methods where assigning the local variable takes a constant fraction of the overall number of operations)

Also: Which version of Java are you using? I cannot replicate these findings. For me, all three are the same timing:
```
IterationBenchmark.forEach            100000  avgt    4  12461.602 ± 438.815  ns/op
IterationBenchmark.forEachCacheArray  100000  avgt    4  12404.329 ± 242.271  ns/op
IterationBenchmark.forEachCacheSize   100000  avgt    4  12649.409 ± 985.465  ns/op
```

> opcodes

All these assumptions are assumptions. The bytecode is not at all what is actually produced by the JIT, which in turn is not at all what the actual operations executed on the CPU are (see pipelining, speculative execution / branch prediction, etc.). Such constant-factor optimizations need (as realistic as possible) profiling as justification.

(regarding point 2: The field however already exists on the heap, so there is no need to write and then free that memory at all!)

> using the final modifier to declare local variables is to conform to the existing code style of fastutil.

Oh, right! Then making all effectively final variables final definitely is good :)

incaseoftrouble:
PS: I don't want to sound overly negative! I've just seen way too often that people try to outsmart a compiler, which more often than not actually is counterproductive.

szarnekow:
It's interesting. I ran the very same benchmark out of curiosity on Hotspot 17.0.2 both on Mac M1 and on x86 and see the same improvements as @mouse0w0 when caching the reference to the array and its length. Even for very small data sets.

incaseoftrouble:
@szarnekow @mouse0w0 

Could you try this benchmark variant:
```java

@State(Scope.Benchmark)
@Warmup(time = 2, iterations = 2)
@Measurement(time = 2, iterations = 4)
@Fork(1)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class IterationBenchmark {
    @Param({"100000"})
    private int n;

    private int[] a;

    @Setup
    public void setup() {
        int[] a = new int[n];
        Arrays.setAll(a, IntUnaryOperator.identity());
        this.a = a;
    }

    @Benchmark
    public void forEach(Blackhole blackhole) {
        for (int i = 0; i < n; i++) {
            a[i] += i;
        }
        blackhole.consume(a);
    }

    @Benchmark
    public void forEachCacheArray(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0; i < n; i++) {
            a[i] += i;
        }
        blackhole.consume(a);
    }

    @Benchmark
    public void forEachCacheSize(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0, max = n; i < max; i++) {
            a[i] += i;
        }
        blackhole.consume(a);
    }
}
```
(I just want to eliminate the low probability that it has something to do with the used blackhole implementation)

mouse0w0:
My benchmark result is based on `# VM version: JDK 17.0.7, Java HotSpot(TM) 64-Bit Server VM, 17.0.7+8-LTS-224`.

incaseoftrouble:
Ok this is very surprising. Let me try if I can reproduce with `HotSpot` and, if so, if it is hotspot being slow or openjdk being fast.

mouse0w0:
Run your benchmark variant in `JDK 17.0.7`, the benchmark results are as follows:
```
Benchmark                                (n)  Mode  Cnt      Score     Error  Units
IterationBenchmark.forEach            100000  avgt    4  24642.541 ± 212.961  ns/op
IterationBenchmark.forEachCacheArray  100000  avgt    4  24641.441 ± 106.699  ns/op
IterationBenchmark.forEachCacheSize   100000  avgt    4  24642.589 ±  38.049  ns/op
```

mouse0w0:
I think the reason for this result is that the benchmark variant only consumes the array object itself, rather than consuming the data within the array object.

incaseoftrouble:
> I think the reason for this result is that the benchmark variant only consumes the array object itself, rather than consuming the data within the array object.

But the compiler cannot know that `.consume(a)` does not access the content of the array, thus it cannot skip the for loop.

mouse0w0:
But my benchmark aligns with the function of forEach, obtaining and consuming one element at a time. We need to clarify what the JVM is doing.

mouse0w0:
My JMH version is `1.37`.

incaseoftrouble:
> But my benchmark aligns with the function of forEach, obtaining and consuming one element at a time. We need to clarify what the JVM is doing.

Absolutely. This was just an attempt at pinpointing where the difference comes from.

Let's try this:

I started from an empty folder and ran what the JMH guide suggests:
```
mvn archetype:generate \
  -DinteractiveMode=false \
  -DarchetypeGroupId=org.openjdk.jmh \
  -DarchetypeArtifactId=jmh-java-benchmark-archetype \
  -DgroupId=org.sample \
  -DartifactId=test \
  -Dversion=1.0
```
Then, I put the following code into `src/main/java/bench/IterationBenchmark.java`:
```java
package bench;

import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.infra.Blackhole;


@State(Scope.Thread)
@Warmup(iterations = 3)
@Measurement(iterations = 5)
@Fork(1)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class IterationBenchmark {
    @Param({"100000"})
    private int n;

    private int[] a;

    @Setup
    public void setup() {
        int[] a = new int[n];
        for (int i = 0; i < n; i++) {
            a[i] = i;
        }
        this.a = a;
    }

    @Benchmark
    public void forEach(Blackhole blackhole) {
        for (int i = 0; i < n; i++) {
            blackhole.consume(a[i]);
        }
    }

    @Benchmark
    public void forEachCacheArray(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0; i < n; i++) {
            blackhole.consume(a[i]);
        }
    }

    @Benchmark
    public void forEachCacheSize(Blackhole blackhole) {
        int[] a = this.a;
        for (int i = 0, max = n; i < max; i++) {
            blackhole.consume(a[i]);
        }
    }
}
```
followed by `mvn clean verify` and `java -jar target/benchmarks.jar`.

szarnekow:
Original benchmark; M1 native;

`JDK 19.0.2, OpenJDK 64-Bit Server VM, 19.0.2+7`

```
Benchmark                                   (n)  Mode  Cnt  Score   Error  Units
IterationBenchmark.forEach                   10  avgt   10  7,168 ± 0,177  ns/op
IterationBenchmark.forEachCacheArrayLength   10  avgt   10  2,375 ± 0,129  ns/op
IterationBenchmark.forEachCacheField         10  avgt   10  3,860 ± 0,141  ns/op
```

`JDK 21.0.1, OpenJDK 64-Bit Server VM, 21.0.1+12-LTS`

```
Benchmark                                   (n)  Mode  Cnt  Score   Error  Units
IterationBenchmark.forEach                   10  avgt   10  2,547 ± 0,059  ns/op
IterationBenchmark.forEachCacheArrayLength   10  avgt   10  2,541 ± 0,036  ns/op
IterationBenchmark.forEachCacheField         10  avgt   10  2,556 ± 0,065  ns/op
```

Numbers on 17.0.9 or newer look similar to 21 for me

mouse0w0:
Upgrade JVM to `17.0.10` and get the same results as you:
```
# JMH version: 1.37
# VM version: JDK 17.0.10, Java HotSpot(TM) 64-Bit Server VM, 17.0.10+11-LTS-240
# VM invoker: C:\Program Files\jdk-17\bin\java.exe
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
```
and
```
Benchmark                                (n)  Mode  Cnt      Score     Error  Units
IterationBenchmark.forEach            100000  avgt    5  13633.378 ± 892.284  ns/op
IterationBenchmark.forEachCacheArray  100000  avgt    5  13517.690 ±  47.061  ns/op
IterationBenchmark.forEachCacheSize   100000  avgt    5  13550.632 ± 198.555  ns/op
```

incaseoftrouble:
Huh! I didn't expect this to come out of a minor version change. Well, that at least explains what happens in that particular case ... Thanks for trying with an upgraded version!

mouse0w0:
Benchmark in JVM versions `17.0.8` and `17.0.9` respectively, the results indicate that the change occurred in `17.0.9`.

incaseoftrouble:
Aha! The 17.0.9 changelog includes `hotspot/compiler	C2 Blackholes should allow load optimizations` 

Since your performance was similar when the blackhole was moved outside the loop, I am guessing this is the culprit. This would mean the performance difference was caused by blackhole and would not be observable in real applications (which aligns with what I have seen in e.g. JBDD)

drv/ArrayList.drv
```diff
@@ -450,11 +450,9 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 #else
 		if (wrapped) a = ARRAYS.ensureCapacity(a, capacity, size);
 		else {
-			if (capacity > a.length) {
-				final Object[] t = new Object[capacity];
-				System.arraycopy(a, 0, t, 0, size);
-				a = (KEY_GENERIC_TYPE[])t;
-			}
+			final Object[] t = new Object[capacity];
+			System.arraycopy(a, 0, t, 0, size);
```
incaseoftrouble, line 454: Shouldn't Arrays.copyOf be applicable here? This can potentially be faster by a) calling realloc and b) avoid initializing the new array with nulls.

mouse0w0: `Arrays.copyOf` copies the entire array data rather than copying an array data of the `size` of the list, which may result in performance differences. In the case of copying the entire array data, their time consumption is the same:
```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 2)
@Measurement(iterations = 2)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ArrayCopyBenchmark {
    private int oldCapacity = 1000;
    private int newCapacity = oldCapacity + oldCapacity >> 1;

    private int[] a;

    @Setup
    public void setup() {
        a = new int[oldCapacity];
        Arrays.setAll(a, IntUnaryOperator.identity());
    }


    @Benchmark
    public void grow() {
        int[] t = new int[newCapacity];
        System.arraycopy(a, 0, t, 0, oldCapacity);
        a = t;
    }

    @Benchmark
    public void copyOf() {
        a = Arrays.copyOf(a, newCapacity);
    }
}
```
Output:
```
# JMH version: 1.37
# VM version: JDK 17.0.10, Java HotSpot(TM) 64-Bit Server VM, 17.0.10+11-LTS-240
# VM invoker: C:\Program Files\jdk-17\bin\java.exe
# VM options: -javaagent:C:\Program Files\IntelliJ IDEA 2023.2.5\lib\idea_rt.jar=4477:C:\Program Files\IntelliJ IDEA 2023.2.5\bin -Dfile.encoding=UTF-8
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 2 iterations, 10 s each
# Measurement: 2 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                  Mode  Cnt    Score   Error  Units
ArrayCopyBenchmark.copyOf  avgt    2  239.428          ns/op
ArrayCopyBenchmark.grow    avgt    2  241.533          ns/op
```

incaseoftrouble: > Arrays.copyOf copies the entire array data rather than copying an array data of the size of the list, which may result in performance differences.

note that `new Object[capacity]` needs to fill the array with `null`

But yes, my comment was under the assumption that `size == a.length` which it need not be. I guess it wouldn't make a difference.

mouse0w0: You are right. In cases where `size != a.length`, `Arrays.copyOf` is better.
```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 3)
@Measurement(iterations = 10)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ArrayCopyBenchmark {
    private int size = 50000;
    private int oldCapacity = 100000;
    private int newCapacity = oldCapacity + oldCapacity >> 1;

    private int[] a;

    @Setup
    public void setup() {
        a = new int[oldCapacity];
        Arrays.setAll(a, IntUnaryOperator.identity());
    }


    @Benchmark
    public int[] ensureCapacity() {
        int[] t = new int[newCapacity];
        System.arraycopy(a, 0, t, 0, size);
        a = t;
        return a;
    }

    @Benchmark
    public int[] copyOf() {
        a = Arrays.copyOf(a, newCapacity);
        return a;
    }
}
```
Output:
```
# JMH version: 1.37
# VM version: JDK 17.0.10, Java HotSpot(TM) 64-Bit Server VM, 17.0.10+11-LTS-240
# VM invoker: C:\Program Files\jdk-17\bin\java.exe
# VM options: -javaagent:C:\Program Files\IntelliJ IDEA 2023.2.5\lib\idea_rt.jar=6866:C:\Program Files\IntelliJ IDEA 2023.2.5\bin -Dfile.encoding=UTF-8
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                          Mode  Cnt      Score    Error  Units
ArrayCopyBenchmark.copyOf          avgt   10  26417.144 ± 83.420  ns/op
ArrayCopyBenchmark.ensureCapacity  avgt   10  25251.914 ± 65.468  ns/op
```

vigna:
Wow. That's a great and constructive discussion!

My 2€¢: I welcomed the PRs from @mouse0w0 because this is how fastutil code is written. Beyond 2 accesses to the same array I try to cache. I know it's up and down with actual improvements etc., but semantically it really makes sense and aften works (remember we're supporting Java 8). 

@mouse0w0 noted that in some places I wasn't being coherent and fixed those places. So I'm totally happy to see this discussion but at least for the time being (and seeing different results on different architectures) for 3 accesses or more the rule is to cache. If you spot a relevant method where this does not happen I'm happy to accept PRs.

incaseoftrouble:
> for 3 accesses or more the rule is to cache

I'm fine with that. Its just important to note that this primarily is a style decision, not performance.

The only thing I am "unhappy" about is the `for (int i = 0, max = size; ...)`. I find these not nice to read. If the size field should be copied to local, I would prefer ? int size = this.size; for (int i = 0; i < size; i++) {...}`.

drv/ArrayList.drv
```diff
@@ -819,11 +829,13 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
 		it.unimi.dsi.fastutil.Arrays.ensureFromTo(size, from, to);
+		final KEY_TYPE[] a = this.a;
 		System.arraycopy(a, to, a, from, size - to);
 		size -= (to - from);
 #if KEYS_REFERENCE
-		int i = to - from;
-		while(i-- != 0) a[size + i] = null;
+		for (int i = from; i < to; i++) {
+			a[i] = null;
+		}
```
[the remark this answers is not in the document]
vigna: Yes, I think that'd be a good idea.

vigna:
Well, the rule tries to avoid the bad case. It is physically impossible to try with all type combinations and all instances whether there will be a negative impact on performance. Thus the 3+ rule.

What about for(i = 0, _size = size, i < _size; i++)?

incaseoftrouble:
> Well, the rule tries to avoid the bad case. It is physically impossible to try with all type combinations and all instances whether there will be a negative impact on performance. Thus the 3+ rule.

Sure, I just meant that we cant just assume that local caching makes things faster. We tested it, saw that it doesn't make a difference, great

> for(i = 0, _size = size, i < _size; i++)?

I equally dislike that - for me, a for loop should just have a single initialization. But its your project and your style decision :)

drv/ArrayList.drv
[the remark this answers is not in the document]
mouse0w0: Agree, and it's a mistake, I've fixed it.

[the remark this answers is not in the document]
mouse0w0: Did a benchmark for it:
```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 2)
@Measurement(iterations = 2)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ArrayFillBenchmark {
    @Param({"10", "1000", "100000"})
    private int n;

    private Object[] a;

    @Setup
    public void setup() {
        a = new Object[n];
        Arrays.setAll(a, t -> t);
    }

    @Benchmark
    public Object[] fill() {
        Arrays.fill(a, 0, a.length, null);
        return a;
    }

    @Benchmark
    public Object[] forward() {
        for (int i = 0; i < a.length; i++) {
            a[i] = 1;
        }
        return a;
    }
}
```
Output:
```
# JMH version: 1.37
# VM version: JDK 17.0.10, Java HotSpot(TM) 64-Bit Server VM, 17.0.10+11-LTS-240
# VM invoker: C:\Program Files\jdk-17\bin\java.exe
# VM options: -javaagent:C:\Program Files\IntelliJ IDEA 2023.2.5\lib\idea_rt.jar=10427:C:\Program Files\IntelliJ IDEA 2023.2.5\bin -Dfile.encoding=UTF-8
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 2 iterations, 10 s each
# Measurement: 2 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                      (n)  Mode  Cnt       Score   Error  Units
ArrayFillBenchmark.fill         10  avgt    2       4.254          ns/op
ArrayFillBenchmark.fill       1000  avgt    2     258.657          ns/op
ArrayFillBenchmark.fill     100000  avgt    2   25504.678          ns/op
ArrayFillBenchmark.forward      10  avgt    2      16.514          ns/op
ArrayFillBenchmark.forward    1000  avgt    2    1709.589          ns/op
ArrayFillBenchmark.forward  100000  avgt    2  160617.181          ns/op
```

mouse0w0:
> for (int i = 0, max = size; ...)

This code style has been applied to some of fastutil's templates, and if we don't use it, we'll need to change some of the templates to keep it consistent.

> int size = this.size; for (int i = 0; i < size; i++) {...}

Personally, I prefer this code style as well,  `int size = this.size` is already heavily used in this pull request.

vigna:
> > for (int i = 0, max = size; ...)
> 
> This code style has been applied to some of fastutil's templates, and if we don't use it, we'll need to change some of the templates to keep it consistent.
> 
> > int size = this.size; for (int i = 0; i < size; i++) {...}
> 
> Personally, I prefer this code style as well, `int size = this.size` is already heavily used in this pull request. And to simplify the code in the future, we can simply remove it without any problems.

Ok, so let's go for `int size = this.size;`. One just has to be sure not to change the semantics of the the method, as `this.` is not compulsory in Java. I'd thus suggest `final int size = this.size;`. In this way, if in the rest of the method where was some mutation for `this.size` we'll get an error.

drv/ArrayList.drv
```diff
@@ -662,10 +666,14 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
-					action.accept(a[from + (lastReturned = pos++)]);
+				final int from = SubList.this.from, to = SubList.this.to;
+				for (int i = from + pos; i < to; i++) {
+					action.accept(a[i]);
 				}
+
+				final int max = to - from;
+				pos = max;
+				lastReturned = max - 1;
 			}
```
mouse0w0, line 677: The `forEachRemaining` change is meaningless (maybe slower) in the latest benchmark, and I'm not sure if I should keep it. I did this optimization because Java's `ArrayList` uses the same code, and it seems more efficient (less assignment and self-increment).
Benchmark:
```java
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 3)
@Measurement(iterations = 5)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ForEachRemainingBenchmark {
    @Param({"10", "1000", "100000"})
    private int max;
    private int[] a;

    @Setup
    public void setup() {
        a = new int[max];
        Arrays.setAll(a, IntUnaryOperator.identity());

    }

    private int pos;
    private int last;

    @Benchmark
    public void raw(Blackhole blackhole) {
        pos = max >> 1;
        last = pos - 1;

        while (pos < max) {
            blackhole.consume(a[last = pos++]);
        }
        blackhole.consume(pos);
        blackhole.consume(last);
    }

    @Benchmark
    public void optimize(Blackhole blackhole) {
        pos = max >> 1;
        last = pos - 1;

        for (int i = pos; i < max; i++) {
            blackhole.consume(a[i]);
        }
        pos = max;
        last = max - 1;
        blackhole.consume(pos);
        blackhole.consume(last);
    }
}
```
Output:
```
Benchmark                            (max)  Mode  Cnt     Score    Error  Units
ForEachRemainingBenchmark.optimize      10  avgt    5     3.331 ±  0.018  ns/op
ForEachRemainingBenchmark.optimize    1000  avgt    5    56.202 ±  0.023  ns/op
ForEachRemainingBenchmark.optimize  100000  avgt    5  6028.862 ±  7.591  ns/op
ForEachRemainingBenchmark.raw           10  avgt    5     3.786 ±  0.006  ns/op
ForEachRemainingBenchmark.raw         1000  avgt    5    53.319 ±  0.026  ns/op
ForEachRemainingBenchmark.raw       100000  avgt    5  6007.510 ± 14.774  ns/op
```

vigna: Well, that's a sensible criterion, but the ArrayList code has been written by many people over the years, and optimized for different implementations of Collections and different JVMs, so it's not necessary the best.

However, the difference is so small that I'd keep the new ArrayList-style code. It's definitely cleaner, and writing in Rust I realized that exceeding with side-effects in expression evaluation doesn't do a good service to anybody.

vigna:
This PR makes tests fail (didn't you check?). Apparently there is a change of behavior of the list used to check the behavior of big lists. I don't have time at the moment to chase this, as it might be involved.

mouse0w0:
My fault. I am not yet familiar with using ant junit, and the lack of prominent failures made me ignore it. I will fix it.

vigna:
You can run the test `ObjectBigArrayBigListTest` from your IDE, or `ant junit` but that will run everything.

Note that in principle something that you _fixed_ might have uncovered a bug in big lists. I am not keen to think that because those tests ran happily for +20 years.

Unfortunately the failed test is a stress test that call itself recursively, which makes not easy to track the point in which divergence between the big list and the standard list happens.

mouse0w0:
The issue is located in `Arrays.ensureFromTo`. When calling `clear` on an empty SubList, it calls `ObjectArrayList.removeElement`, with both `from` and `to` being 0. According to the Javadoc, the `to` parameter in `removeElement` is exclusive, but the `to` parameter in `Arrays.ensureFromTo` is inclusive. Upon inspection, the `to` parameters of all methods that call `Arrays.ensureFromTo` is exclusive. Therefore, `Arrays.ensureFromTo` should be fixed.

mouse0w0:
According to the implementation of Java's `Objects.checkFromToIndex`, the range from 0 (inclusive) to 0 (exclusive) is valid, which means it doesn't include any indices. `Arrays.ensureFromTo` does not need to be changed, but its javadoc is incorrect.

vigna:
I'm totally confused. What did you fix codewise (forget the docs for the moment)?

mouse0w0:
The problem lies in my `removeElements` method, where I mistakenly filled the object array. This has been fixed in #327. However, the javadoc description for the `to` parameter in `Arrays.ensureFromTo` is incorrect, which misled me into thinking the issue was in its implementation. The `Arrays.ensureFromTo`  original javadoc is as follows:
https://github.com/vigna/fastutil/blob/5eaffbbd849ec8b034046e4a8b1a0cc410457778/src/it/unimi/dsi/fastutil/Arrays.java#L52-L53
https://github.com/vigna/fastutil/blob/5eaffbbd849ec8b034046e4a8b1a0cc410457778/src/it/unimi/dsi/fastutil/Arrays.java#L63-L65
It should be:
<code>	 * @param to an end index (~~inclusive~~ exclusive).</code>

vigna:
Ahh OK, I see. The text is correct. The parameter description is wrong.

vigna:
WTF, I have the new release but Sonatype gives me Unauthorized (401). Nothing has changed in my configuration since 5 months ago. If someone has suggestions, let me know, I tried on the community forum but I just met other people with the same problem...

mouse0w0:
Sonatype now enforces token authentication instead of username and password. Refer to the user manual to generate your token.

vigna:
Thanks for the tip! They could provide a new sensible error message, frankly (or I'm just getting addicted to the Rust ecosystem).

drv/ArrayList.drv
mouse0w0: I made a silly mistake here. `Arrays.copyOf` is actually slower, so it's the worse choice.

PR #327, opened 2024-07-22
Optimize ArrayList

mouse0w0:
Previous #316

### DIFF
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -448,7 +448,11 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		a = ARRAYS.ensureCapacity(a, capacity, size);
 #else
 		if (wrapped) a = ARRAYS.ensureCapacity(a, capacity, size);
-		else a = Arrays.copyOf(a, capacity);
+		else {
+			final Object[] t = new Object[capacity];
+			System.arraycopy(a, 0, t, 0, size);
+			a = (KEY_GENERIC_TYPE[])t;
+		}
 #endif
 		assert size <= a.length;
 	}
@@ -467,7 +471,11 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		a = ARRAYS.forceCapacity(a, capacity, size);
 #else
 		if (wrapped) a = ARRAYS.forceCapacity(a, capacity, size);
-		else a = Arrays.copyOf(a, capacity);
+		else {
+			final Object[] t = new Object[capacity];
+			System.arraycopy(a, 0, t, 0, size);
+			a = (KEY_GENERIC_TYPE[])t;
+		}
 #endif
 		assert size <= a.length;
 	}
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -835,10 +835,11 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
 		it.unimi.dsi.fastutil.Arrays.ensureFromTo(size, from, to);
 		final KEY_TYPE[] a = this.a;
 		System.arraycopy(a, to, a, from, size - to);
+		final int newSize = size - (to - from);
 #if KEYS_REFERENCE
-		Arrays.fill(a, to, size, null);
+		for (int i = newSize; i < size; i++) a[i] = null;
 #endif
-		this.size -= (to - from);
+		this.size = newSize;
 	}
 
 
--- a/drv/ArrayList.drv
+++ b/drv/ArrayList.drv
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2002-2024 Sebastiano Vigna
+ * Copyright (C) 2002-2025 Sebastiano Vigna
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jmr:
@mouse0w0 @vigna What's the status of this?

vigna:
What I wrote.

drv/ArrayList.drv
```diff
@@ -442,20 +442,13 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
 	 * @param capacity the new minimum capacity for this array list.
 	 */
-	SUPPRESS_WARNINGS_KEY_UNCHECKED
 	public void ensureCapacity(final int capacity) {
 		if (capacity <= a.length || (a == ARRAYS.DEFAULT_EMPTY_ARRAY && capacity <= DEFAULT_INITIAL_CAPACITY)) return;
 #if KEYS_PRIMITIVE
 		a = ARRAYS.ensureCapacity(a, capacity, size);
 #else
 		if (wrapped) a = ARRAYS.ensureCapacity(a, capacity, size);
-		else {
```
vigna, line 452: How can this be equivalent? There's no test.

mouse0w0: You're right. I made a silly mistake in [this part](https://github.com/vigna/fastutil/pull/316/files/ed0c36222015f598e0874f39767c11705811b94d#r1487410614).

drv/ArrayList.drv
```diff
@@ -662,10 +674,14 @@ public class ARRAY_LIST KEY_GENERIC extends ABSTRACT_LIST KEY_GENERIC implements
...
-				while(pos < max) {
-					action.accept(a[from + (lastReturned = pos++)]);
+				final int from = SubList.this.from, to = SubList.this.to;
+				for (int i = from + pos; i < to; i++) {
+					action.accept(a[i]);
 				}
+
+				final int max = to - from;
+				pos = max;
+				lastReturned = max - 1;
```
mouse0w0, line 684: Should this change be reverted? [There is no significant performance difference between the two implementations](https://github.com/vigna/fastutil/pull/316#discussion_r1493798878), but the original version allowed consumers to keep calling Iterator’s `next` method in `forEachRemaining, whereas the new implementation follows the style of Java’s ArrayList.

mouse0w0: Benchmark:
```java
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 3)
@Measurement(iterations = 5)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class ForEachRemainingBenchmark {
    @Param({"10", "1000", "100000"})
    private int max;
    private int from;
    private int to;
    private int[] a;

    @Setup
    public void setup() {
        a = new int[max];
        Arrays.setAll(a, IntUnaryOperator.identity());
        from = max >> 2;
        to = max - from;
    }

    private int pos;
    private int last;

    @Benchmark
    public void old(Blackhole blackhole) {
        pos = 0;
        last = pos - 1;

        final int[] a = this.a;
        final int max = to - from;
        while (pos < max) {
            blackhole.consume(a[from + (last = pos++)]);
        }
        blackhole.consume(pos);
        blackhole.consume(last);
    }

    @Benchmark
    public void now(Blackhole blackhole) {
        pos = 0;
        last = pos - 1;

        final int[] a = this.a;
        final int from = this.from, to = this.to;
        for (int i = from + pos; i < to; i++) {
            blackhole.consume(a[i]);
        }

        final int max = to - from;
        pos = max;
        last = max - 1;
        blackhole.consume(pos);
        blackhole.consume(last);
    }
}
```
Output:
```
# JMH version: 1.37
# VM version: JDK 21.0.7, OpenJDK 64-Bit Server VM, 21.0.7+6-LTS
# VM invoker: C:\Program Files\Zulu\zulu-21\bin\java.exe
# VM options: -javaagent:C:\Program Files\JetBrains\IntelliJ IDEA 2025.1.1.1\lib\idea_rt.jar=3441 -Dfile.encoding=UTF-8 -Dsun.stdout.encoding=UTF-8 -Dsun.stderr.encoding=UTF-8
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                       (max)  Mode  Cnt     Score    Error  Units
ForEachRemainingBenchmark.now      10  avgt    5     2.427 ±  0.052  ns/op
ForEachRemainingBenchmark.now    1000  avgt    5    38.464 ±  0.511  ns/op
ForEachRemainingBenchmark.now  100000  avgt    5  5744.566 ± 63.299  ns/op
ForEachRemainingBenchmark.old      10  avgt    5     5.107 ±  0.327  ns/op
ForEachRemainingBenchmark.old    1000  avgt    5    65.604 ±  0.722  ns/op
ForEachRemainingBenchmark.old  100000  avgt    5  6414.992 ± 77.098  ns/op
```
The latest benchmark results show that the current implementation is faster, but we should be careful about how different implementations may affect the method's behavior.

drv/ArrayList.drv
vigna: Please re-review carefully all your changes, it's a lot of code and I just spotted this—since it has been a while, you might spot other mistakes, too.

drv/ArrayList.drv
vigna: Yeah, I would revert the change.

mouse0w0:
I realized that these changes have no effect on modern JVMs. It's a bit disappointing, but I'm leaving it here for reference.

PR #330, opened 2024-08-13
Add ofEntries static method to Map interfaces

### LINKED ISSUE #329: Map.ofEntries equivalent for fastutil maps
Adding something like [Map#ofEntries](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#ofEntries(java.util.Map.Entry...)) to the fastutil maps would be neat in my humble opinion.
Right now you gotta do ugly stuff like `Int2ObjectMaps.unmodifiable(new Int2ObjectOpenHashMap(Map.ofEntries(...)))`
I could make a PR if you don't have time to.
(btw great library much love ❤️)

viciscat:
Pretty simple PR, I think
Things to note:
- Java doc is a bit quaint because I thought it was useless to copy paste the original methods
- an `entry(key, value)` static method was also added to make it easier to use, just like `java.util.Map`
- I did not include `Int2IntMapTest` because I would have put the same things I did in `Object2ObjectMapTest`
- More tests could probably be added but I have no idea what they could be

Close #329

### DIFF
--- a/drv/Map.drv
+++ b/drv/Map.drv
@@ -823,6 +823,41 @@ public interface MAP KEY_VALUE_GENERIC extends FUNCTION KEY_VALUE_GENERIC, Map<K
 
 #endif
 
+	/** Returns an unmodifiable, type-specific map with the specified entries.
+	 * See {@link java.util.Map#ofEntries(Map.Entry[])} */
+#if !KEYS_PRIMITIVE || !VALUES_PRIMITIVE
+	@SafeVarargs
+	@SuppressWarnings({"varargs", "unchecked"})
+#else
+	@SuppressWarnings("varargs")
+#endif
+	static KEY_VALUE_GENERIC MAP KEY_VALUE_GENERIC ofEntries(MAP.Entry KEY_VALUE_GENERIC ... entries) {
+		if (entries.length == 0) return MAPS.EMPTY_MAP;
+		if (entries.length == 1) return MAPS.singleton(entries[0].ENTRY_GET_KEY(), entries[0].ENTRY_GET_VALUE());
+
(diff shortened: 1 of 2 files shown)


drv/Map.drv
```diff
@@ -823,6 +823,41 @@ public interface MAP KEY_VALUE_GENERIC extends FUNCTION KEY_VALUE_GENERIC, Map<K
...
+			MAP.Entry KEY_VALUE_GENERIC entry = entries[i];
+			keys[i] = entry.ENTRY_GET_KEY();
+			// Prevent duplicate keys
+			for (int j = 0; j < i; ++j) if (KEY_EQUALS(keys[j], keys[i])) throw new IllegalArgumentException("duplicate key: " + keys[i]);
+			vals[i] = entry.ENTRY_GET_VALUE();
+		}
+		if (entries.length <= 8) {
+			return MAPS.unmodifiable(new ARRAY_MAP KEY_VALUE_GENERIC_DIAMOND(keys, vals, entries.length));
+		} else {
+			return MAPS.unmodifiable(new OPEN_HASH_MAP KEY_VALUE_GENERIC_DIAMOND(keys, vals, 0.75f));
```
incaseoftrouble, line 851: this creates the keys / values array unnecessarily. I think the code should be like
```
if (number of entries is small)
  create keys and vals as above (with duplicate checking)
  return array map with keys and vales
else
  create appropriately sized hash map
  for each entry:
    if map.put(entry.key, entry.value) != null then throw exception
  return hash map
```

viciscat: you are quite right, let me do that

viciscat: would using `contains` instead of the return value of `#put` be better? For the edge case where the previous value is `null`

Issue #336, opened 2024-11-18
Feature Request: Implement Object2DoubleOpenHashMap.mergeDouble to avoid double-hashing


mhansen:
Hi, I was recently optimising some hashmap heavy code. We got a big speedup from moving from `HashMap<Object, Double>` to `Object2DoubleOpenHashMap`, so first let me say thank you for fastutil, it's great.

We noticed an opportunity in the profiling afterwards though. I expected [`mergeDouble`](https://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/doubles/Double2DoubleMap.html#mergeDouble(double,double,it.unimi.dsi.fastutil.doubles.DoubleBinaryOperator)) to have an optimisation to hash the input once, then find the position of the data, then replace the data in that position. Like `HashMap.compute`.

But in the profile I see that `mergeDouble` is implemented only in the interface as `Object2DoubleMap.mergeDouble`, which doesn't know about hashing, and is implemented in terms of calling `double getDouble(Object)` then `put(Object, double)` This ends up hashing the `Object` twice, and finding the hashmap position twice (including calling `.equals` to see if it's the right position).

It's not a big deal or anything; we're more than happy with the performance improvements delivered. But I thought I'd report it in case it's an idea you like. Here's a flamegraph:

![image](https://github.com/user-attachments/assets/b58a91ae-faba-478a-8036-b8199c26856d)

vigna:
Mmmmh. There is a merge implementation for hash maps, so I'm wondering why it isn't called. Don't you see it in a stack trace? In any case, that wouldn't solve your problem, in the sense that fastutil internal functions find and insert do not carry the hash. You must understand fastutil was designed more than 20 years ago, when compound operations such as merge and compute were not in the interface. It is right that in the present situation it would be better to have find and insert accepting a hash, so that compound operations compute it just once. It's a good suggestion, I just don't know when I'll find the bandwidth to implement it (but your are free to send PRs 😂; just joking, I know it's complex code).

mhansen:
Thank you for this background!

I'll go off what I see in the javadoc as it's a little hard to link to the generated code.

> There is a merge implementation for hash maps, so I'm wondering why it isn't called. Don't you see it in a stack trace?

[...]

PR #337, opened 2024-11-19
OpenHashMaps.mergePRIMITIVE: Avoid double-hashing key

mhansen:
The `#if VALUES_PRIMITIVE && ! VALUE_CLASS_Boolean` comes from the superclass in Map.drv: https://github.com/vigna/fastutil/blob/355d8ebfaf2bcc4032cc1c25bec9f3fe827f8175/drv/Map.drv#L649

The implementation mostly is copied from the method below (`merge`), with dead code removed because `VALUES_PRIMITIVE` is always true.

I'm using the `METHOD_ARG_VALUE_BINARY_OPERATOR` so that we prefer using the JDK ${Primitive}BinaryOperator classes, and fall back to the fastutil primitive-BinaryOperators where not present in the JDK.

I ran the junit tests locally and they pass.

Towards #336

### DIFF
--- a/drv/OpenHashMap.drv
+++ b/drv/OpenHashMap.drv
@@ -1253,6 +1253,25 @@ public class OPEN_HASH_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENE
 		return value[pos] = newVal;
 	}
 
+#if VALUES_PRIMITIVE && ! VALUE_CLASS_Boolean
+	/** {@inheritDoc} */
+	@Override
+	public VALUE_GENERIC_TYPE MERGE_VALUE(final KEY_GENERIC_TYPE k, final VALUE_GENERIC_TYPE v, METHOD_ARG_VALUE_BINARY_OPERATOR remappingFunction) {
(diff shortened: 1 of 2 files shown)


vigna:
BTW, did you check that your code is actually executed (either by prints or by coverage tools)? The dispatch of that stuff is absolutely devilish...

mhansen:
Thank you! You're right, I should check this is actually executing. I think I'll write a test with an object that counts how many times it was hashed, asserting it was just hashed once after.

vigna:
You can also just use a coverage tool. Or add a print in the code and make a one-off test, checking that something is printed.

mhansen:
I've confirmed it's executing. It's actually down from 3 hashes (get, containsKey, put) down to 1. I'm pretty happy to write a test for this; fastutil is a long-term project and it seems worth adding some test to stop this unintentionally regressing in future.

mhansen:
I think I can add a test in `Object2IntOpenHashMapTest`.

mhansen:
PTAL; I've added a test in commit 1, then reduced the allocations in commit 2.

PR #352, opened 2025-06-04
.gitignore: Add reports/

jmr:
Ignore test output.


jmr:
I was wondering why this was closed, but now I see it was part of 934f17e (Tweaking).

PR #353, opened 2025-06-04
Use static hashCode() methods

jmr:
Java 8 added static hashCode() methods like Double.hashCode(double).

https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#hashCode-double-

These are equivalent to what HashCommon.double2int(double) claims to do

https://javadoc.io/doc/it.unimi.dsi/fastutil/latest/it/unimi/dsi/fastutil/HashCommon.html#double2int(double)

but double2int() uses Double.doubleToRawLongBits(), whereas Double.HashCode() uses Double.doubleToLongBits().  Therefore, also update KEY2INTHASH() and KEY2LONGHASH() to use doubleToLongBits().

Since HashCommon is a public class, I didn't delete the old HashCommon.double2int(), float2int(), and long2int() functions.

### DIFF
--- a/gencsource.sh
+++ b/gencsource.sh
@@ -874,17 +874,17 @@ $(if [[ "${CLASS[$k]}" != "" && "${CLASS[$v]}" != "" ]]; then\
 "#else\n"\
 \
 "#if KEY_CLASS_Float\n"\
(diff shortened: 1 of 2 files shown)


vigna:
You might be right. Consider that the code was written 30 years ago. If you're right, it should be possible to write a test in which a map containing different types of NaN misbehaves.

PR #355, opened 2025-06-21
FastByteArrayOutputStream: Small Usability improvement 

magicprinc:
1) copyright 2024→2025
2) toByteArray similar to ByteArrayOutputStream#toByteArray() for easy migration and daily life
3) super class close() and write(byte[]) repeated without IOException

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ pom-*.xml
 .classpath
 .project
 .ant-targets-build.xml
+.idea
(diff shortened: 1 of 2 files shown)


vigna:
I merged, but 1) no check for null in write—it's not like that in the parent class, and 2) toByteArray() always copies the array, as ByteArrayOutputStream does. You are even referring to its documentation—you cannot have a different behavior.

magicprinc:
Thank you! 🥳

Yeah, I tend to over-protect, over-care and over-optimize 😥

PR #369, opened 2025-09-19
Type specific Map#foreach

viciscat:
Title says it all.

I also tried doing a custom implementation of `forEach` in `OpenHashMap` to remove that one allocation of `MapEntry` but I don't know if that's wanted.
Didn't do many javadoc.

### DIFF
--- a/drv/OpenHashMap.drv
+++ b/drv/OpenHashMap.drv
@@ -1333,6 +1333,7 @@ public class OPEN_HASH_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENE
 		return size == 0;
 	}
 
+#ifndef Linked
 	@Override
 	public void forEach(final EntryConsumer KEY_VALUE_GENERIC consumer) {
 		if (containsNullKey) {
@@ -1344,7 +1345,7 @@ public class OPEN_HASH_MAP KEY_VALUE_GENERIC extends ABSTRACT_MAP KEY_VALUE_GENE
 				consumer.accept(key[pos], value[pos]);
 			}
 	}
-
+#endif
 
 	/** The entry class for a hash map does not record key and value, but
 	 * rather the position in the hash table of the corresponding entry. This


vigna:
Oh well. Which things?

viciscat:
doing a custom implementation of `forEach` for the other maps like what I did in `OpenHashMap` and better javadoc I guess? pretty sure I didn't put any on the type specific `forEach`

vigna:
> I also tried doing a custom implementation of `forEach` in `OpenHashMap` to remove that one allocation of `MapEntry` but I don't know if that's wanted. Didn't do many javadoc.

BTW the fast iterators for hash maps do exactly that—they reuse the entry.

viciscat:
> BTW the fast iterators for hash maps do exactly that—they reuse the entry.

You see the custom implementation removes one (1) whole allocation. Since it doesn't allocate the entry once! Wow such optimization such performance

vigna:
Ok, admittedly I've been very lazy in reviewing this PR.

First, we have a Consumer type-specific impl, so I'll move the EntryConsumer thing to its own file. It doesn't make any sense to have a more specific thing for `forEach`.

Then I'll have a look into the docs.

viciscat:
Unless I'm stupid there's no type-specific BiConsumer, I could have made them their own classes but I did not want to add 64 new files

viciscat:
awesome, can't wait for type-specific BiFunctions next 🙃

vigna:
That's why it's not likely to happen 😬

vigna:
Ok, I have revised everything. I'd like to ask you to help me with a couple of things:

- The implementation of `forEach` for `OpenHashMap.drv` works only for the non-linked case: is there a particular reason?
- What about implementations for other maps?

viciscat:
- Didn't notice Linked was in the same file and the implementation doesn't work follow the correct order using `link` array. I didn't have time to do a proper implementation so I just quickly put a `#ifndef Linked` to not have a wrong implementation on that one.
- As said in the message, I wasn't sure if this was a welcome addition, so I only did one as a "proof of concept" and made the pull request a draft.

vigna:
I don't even know what a "draft" PR is 😂. Well, you're welcome to do a proper implementation. Even for other maps if you think it's useful.

